PR #43, opened 2018-09-25
Adding install and readme into MAM

EricBoucher22:


EricBoix:
Oh well, let us accept that...

PR #99, opened 2019-07-09
Feature/groups

Crejak:
- Added helper functions to work with groups in 3DTiles tiles.
  - `createTileGroups` creates new tile groups from the given materials and vertex ranges.
  - `createTileGroupsFromBatchIDs` does the same with materials and batch IDs (instead of vertex ranges).
- Along with that, a new form has been added to 3DTilesDebug to show the possibility of the new functions. 
- Weird CSS rules in DocToValidateStyle have been removed / updated

### DIFF
--- a/UDV-Core/src/Utils/3DTiles/3DTilesUtils.js
+++ b/UDV-Core/src/Utils/3DTiles/3DTilesUtils.js
@@ -168,6 +168,132 @@ export function setTileVerticesColor(tile, newColor, indexArray = null) {
 }
 
 /**
+ * Creates tile groups.
+ * 
+ * @param {*} tile The 3DTiles tile.
+ * @param {*} groupColor The color for the specified vertices.
+ * @param {number} groupOpacity The opacity for the specified verticies. It must
+ * be a number between 0 and 1.
+ * @param {Array<Array<number>>} ranges An array of ranges. A range is an array
+ * of size two in the form [start, count], where start is the index of the first
+ * vertex of the range and count is the number of vertices in the range.
+ */
+export function createTileGroups(tile, groupColor = 0xff00ff, groupOpacity = 1, ranges = []) {
+  if (!tile) {
+    throw 'Tile not loaded in view';
+  }
+
+  //Find the 'Mesh' part of the tile
+  while (!!tile.children[0] && !(tile.type === 'Mesh')) {
+    tile = tile.children[0];
+  }
+
+  if (!tile.geometry.attributes._BATCHID) {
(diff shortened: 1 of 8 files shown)


UDV-Core/src/Utils/3DTiles/3DTilesUtils.js
```diff
@@ -167,6 +167,191 @@ export function setTileVerticesColor(tile, newColor, indexArray = null) {
...
+ * @param {Array<any>} materials An array of material parameters. Each entry
+ * should be an object that will be passed as parameter in the THREE.Material
+ * constructor. Accepted values are for example `color` or `opacity` (see
+ * THREE.js documentation).
+ * @param {Array<any>} ranges An array of ranges. A range is an object with 3
+ * propeties :
+ * - `start`: the start index of the group of vertices
+ * - `count`: the number of vertices of the group
+ * - `material`: the index of the material in the materials array
+ */
```
jailln: Could you add an @example similar to the one of createTileGroupsFromBatchIDs() please?

UDV-Core/src/Utils/3DTiles/3DTilesUtils.js
```diff
@@ -167,6 +167,191 @@ export function setTileVerticesColor(tile, newColor, indexArray = null) {
...
+export function createTileGroups(tile, materials, ranges) {
```
jailln: Could you rename materials to something like materialsProperties please to avoid confusion with actual threejs materials and with the materials of the mesh ?

Could you also add a list of the accepted properties please ?

UDV-Core/src/Utils/3DTiles/3DTilesUtils.js
Crejak: Done

UDV-Core/src/Utils/3DTiles/3DTilesUtils.js
Crejak: I renamed it `materialsProps` to keep a short name.

I didn't add the full list of accepted properties because it would be too long. The purpose of this object is to be passed as parameter for the `MeshLambertMaterial` constructor. According to the doc :

> Any property of the material (including any property inherited from Material) can be passed in here.

Instead, I added a comment to explain the above and a link to the relevant documentation (https://threejs.org/docs/#api/en/materials/MeshLambertMaterial).

jailln: Perfect, thanks

PR #117, opened 2019-09-06
Add a link to user tutorial in the help module

jailln:


EricBoix:
I recall we placed a link to the source code of the key-bindings in order to keep track of possible changes (a soft version of [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself)). But that link was broken anyhow, so we can get rid of it...

PR #121, opened 2019-12-09
Added camera position property

clementcolin:
Use extent center by default, else use property camera : position : x,y on the config file


EricBoix:
@clementcolin  Maybe this PR should also contribute a concrete example of usage of this new feature by providing the example you are working on e.g. with the limonestConfig.js file ?

clementcolin:
> 
> 
> @clementcolin Maybe this PR should also contribute a concrete example of usage of this new feature by providing the example you are working on e.g. with the limonestConfig.js file ?

I'll provide an example when i'll push the limonest's example online. Else, the example will not work, since it is still locally hosted

PR #125, opened 2019-12-16
Fixed demo limonest config

clementcolin:
The work on the camera and the demonstration on Limonest now respect the config template


EricBoix:
Fixes as suggested by VJA.

PR #129, opened 2020-01-14
fix new value for heightmap

clementcolin:
As the heightmap was fixed by the Grand Lyon, the min and max value needed to be changed too.


jailln:
Could you add a comment to explain how to obtain these values please ?

By the way; would it be possible to compute them dynamically or to retrieve them from the server ?

clementcolin:
I got them from the Grand Lyon, unfortunately I don't know how they got it. 
I also checked the retrieved data from the WMS source, and there is no information about those values

jailln:
Thanks for checking that out. Could you add a comment giving a link to where you find them and indicating that we don't know yet how to obtain them dynamically please ?

clementcolin:
I'm waiting on a response from the GL staff to know how we can retrieve them from the metadata of the WMS source.
FYI: the heightmap is clamped between 0 and 255, the min and max altitudes is usefull to compute the real altitude of a given point.

PR #130, opened 2020-01-27
New demonstrations - Point clouds - Bron

clementcolin:
Add new demonstration of Bron
Add new demonstration of visualization of point clouds. You can find the note to compute the data set from LAZ file [here](https://github.com/VCityTeam/VCity/wiki/Nuage-de-points)
Add the possibility to change the size of the points through the DemoConfig.json

### DIFF
--- a/UD-Viz-Core/examples/DemoBron/Demo.html
+++ b/UD-Viz-Core/examples/DemoBron/Demo.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>UDV-Core</title>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <link rel="stylesheet" href="../../src/Utils/BaseDemo/css/Demo.css">
+    </head>
+    <body>
+        <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
+        <script src="../../node_modules/moment/moment.js"></script>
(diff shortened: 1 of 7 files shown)


jailln:
* I tested the demos on my computer and the tileset is not displayed; is it working for you ? The tilesets are however accessible from rict2, for instance at http://rict2.liris.cnrs.fr/DataStore/bron_tiles/tileset.json for Bron. Maybe this issue is related to visible ports of rict2 ?

* Regarding the Bron demo:
   * The documents module is connected to a data base with documents representing fourviere and bellecour which are outside the extent specified in this demo. You may need to set up a new server for this demo with documents of Bron.

* Regarding the point clouds demo: 
   * Please update the tileset on rict2 (the config currently points to a local tileset)

* Regarding [the point cloud data set computation](https://github.com/VCityTeam/VCity/wiki/Nuage-de-points): could you translate these notes and add them to UD-Reproducibility please ? (e.g. in the Computations folder)

clementcolin:
* This issue may be related to visible ports indeed, as the current Limonest demo is working fine with the same configuration (data fetch on rict2). I guess that we are not able today to fetch data from rict2 locally. Still, the demo should work fine on the server.

* I launched a UD-Server on rict2 and updated the link to it, it is now separated from the current Lyon demo. For the futur, we may need to be able to select from different UD-Server for each demonstration as we currently can only set up one server on the generalDemoConfig file.

* The tileset is now on rict2
  
* [The documentation](https://github.com/VCityTeam/UD-Reproducibility/tree/master/Computations/ComputePointCloudsLyon) is now on UD-Reproducibility

PR #131, opened 2020-01-28
Add the possibility to chose the document server on the demoConfigFile

clementcolin:
The demoConfigFile now have a field "servers" (was "server") in which we can declare multiple server settings.  

Also, the multilayer data config was missing a property

### DIFF
--- a/UD-Viz-Core/examples/DemoBron/Demo.js
+++ b/UD-Viz-Core/examples/DemoBron/Demo.js
@@ -25,6 +25,8 @@ baseDemo.loadConfigFile('../data/config/generalDemoConfig.json').then(() => {
     const help  = new udvcore.HelpWindow();
     baseDemo.addModuleView('help', help);
 
+    baseDemo.config.server = baseDemo.config.servers["limonest_bron"];   
+
     ////// AUTHENTICATION MODULE
     const authenticationService =
         new udvcore.AuthenticationService(requestService, baseDemo.config);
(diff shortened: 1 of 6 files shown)


UD-Viz-Core/examples/DemoBron/Demo.js
```diff
@@ -25,6 +25,8 @@ baseDemo.loadConfigFile('../data/config/generalDemoConfig.json').then(() => {
     const help  = new udvcore.HelpWindow();
     baseDemo.addModuleView('help', help);
 
+    baseDemo.config.server = baseDemo.config.servers["limonest_bron"];
```
jailln: Could you instead pass `baseDemo.config.servers["limonest_bron"]` (or the equivalent for other demos) directly to the modules that need the server configuration (authentication, documents, contribute, validation, document comments, guided tours and links etc.) since these modules don't seem to need other parts of the config file ?

UD-Viz-Core/examples/DemoStable/Demo.js
```diff
@@ -20,6 +19,9 @@ baseDemo.loadConfigFile('../data/config/generalDemoConfig.json').then(() => {
     ////// REQUEST SERVICE
     const requestService = new udvcore.RequestService();
 
+    baseDemo.config.server = baseDemo.config.servers["lyon"];   
+    baseDemo.config.server.url = baseDemo.config.server.stableUrl;
```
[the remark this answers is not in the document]
clementcolin: Done

UD-Viz-Core/examples/DemoBron/Demo.js
clementcolin: Do you mean something like that 
 *     new udvcore.AuthenticationService(requestService, baseDemo.config.servers["lyon"]);
Actually, I tried it and it seems that the link module needs to have access to all of the configuration file and I found no work around. Also, it would imply that we'll need to specify for each module the server needed.

jailln: Ok, we'll leave it this way then

PR #133, opened 2020-04-07
Udv cpov first demo

ChataingT:
/!\ First test of PR :
On my repository i created a new branch based on UDV-temporal (most recent work of VJA) and name as UDV-CPOV.
Inside i put the demo i have done.
 * Add a module GraphTemporal based on the temporal module
 * Add dependency to display graph with viz.js

The demo created is accessible through http://localhost:8080/examples/DemoGraphTemporal/Demo.html after running locally a 3D-tile-server on http://localhost:8003/ and is not complete : 
 * The graph display is only shown as exemple (fake data hard-coded for demonstration)
 * No interaction between the graph and the map are available
 
Through the PR will the branch UDV-CPOV will be merge on UDV-temporal ?
If yes, i don't think the PR should be accepted.

Should we create the branch UDV-CPOV in vcity/ud-viz and the PR on it ?

### DIFF
--- a/UDV-Core/examples/DemoGraphTemporal/Demo.html
+++ b/UDV-Core/examples/DemoGraphTemporal/Demo.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>UDV-Core</title>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="../../src/Utils/BaseDemo/css/Demo.css">
+</head>
+<body>
+<script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.0/jquery.min.js"></script>
+<script src="../../dist/udvcore.js"></script>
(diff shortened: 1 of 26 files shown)


jailln:
Regarding your question, IMO we should not merge your branch in the master branch since it will also merge my work of the UDV-Temporal branch into the master.

I think we should either:
- Merge your branch UDV-CPOV in UDV-temporal on a regular basis, which in some way makes UDV-temporal your reference branch
- Keep UDV-CPOV as a clean and as your reference branch and create new ones based on UDV-CPOV when you need to implement new features (e.g. UDV-CPOV-cool_feature1). Then you can create PRs to merge your new features in UDV-CPOV (e.g. UDV-CPOV-cool_feature1 into UDV-CPOV), which still allows us to review your work regularly.

The second solution is a bit more complicated but has the advantage to separate the concerns from my work on UDV-Temporal and from your work on graph visualisation; which may allow easier cleaning afterwards. @EricBoix WDYT ?

jailln:
Closing this PR since this work has been enhanced on [this branch](https://github.com/ChataingT/UD-Viz/tree/UD-CPOV_demo)

PR #196, opened 2021-07-19
New examples

clementcolin:
Added unique examples for using widgets independently. All demonstrations (using  other data, server or services) must have their own repository, following the  UD-Viz-Template repository


valentinMachado:
I've repaired this branch with the new api. Everything is working except this example:
examples/MultiLayer/example.html (errors are throwed in the console)

Every demo are the same except the config file, I think it's a bit confusing for new users (I know I did it this way...)
Maybe for these examples we should not use the start method of the udv.Templates.AllWidget but manually adding each widget like it was done before.

In addition every config file should be in the assets folder, since .json are consider as assets data.

EricBoix:
@valentinMachado I'm not sure to understand the branching workflow. It seems that many commits on this branch do not concern the examples but documentation or gaming related work. Besides a branch with 271 changed files is really too hard/long to review. I believe we shouldn't merge this branch as is. It would be clearer simpler to separate concerns and split this branch into at least two different feature branches.

Aditionnally, merge commits from master (refer eg to commit [2a53c9603d59fd19](https://github.com/VCityTeam/UD-Viz/commit/2a53c9603d59fd192e966bd7b5b3e85221cabddc)) should be discouraged in favor of branch rebase.

valentinMachado:
I try to resolve conflict while rebasing this branch with master.

This examples is throwing errors => examples/MultiLayer/example.html
All the rest seem to work well !
Should I merge ?

clementcolin:
Yes please, I'll fix further errors directly on master since it should be minor adjustments (such as links to tilesets)

valentinMachado (approved):

PR #203, opened 2021-09-01
Focus on tileset

clementcolin:
In this PR: 
- new feature to focus automatically on a loaded tileset
- changed the html writing of LayerChoice, using Jquery properly


jailln:
It could be interesting to push the focus on tileset directly in itowns :)

valentinMachado:
@clementcolin what do you think ?

valentinMachado (approved):

Issue #181, opened 2021-06-10
Reconcile prettier and eslint


EricBoix:
Refer to [this comparison](https://stackoverflow.com/questions/44690308/whats-the-difference-between-prettier-eslint-eslint-plugin-prettier-and-eslint)

mathieuLivebardon:
- We should use this package https://github.com/prettier/prettier-eslint with a .prettierrc and .eslintrc 
- We need more rules than those that already exist in .eslintrc, but also for prettier because eslint only looks at Javascript

DiegoVinasco:
this is fixed in IMUV

PR #223, opened 2021-09-13
Update landing page

valentinMachado:
Help widget can be customizable with html files
Landing page has been updated
related to this issue https://github.com/VCityTeam/UD-Viz-docker/issues/1


index.html
```diff
@@ -1,145 +1,162 @@
...
+          </div>
+        </div>
+      </div>
+      <div id="viewer">
+        <div id="projects">
+          <a
+            href="./examples/AllWidget/example.html"
+            target="_blank"
+            rel="noopener"
+            title="This demo implements all widgets and extensions currently available. As of today, this includes :
```
EricBoix, line 51: Some example descriptions use the term "demo" when others use the term "example".
Use the term "example" everywhere.

src/Templates/AllWidget/AllWidget.js
```diff
@@ -35,17 +35,17 @@ export class AllWidget {
   }
 
   start(path) {
-    return new Promise( resolve => {
+    return new Promise((resolve) => {
```
EricBoix, line 38: It seems this file only has coding style changes. Are those changes due to fact that prior to this version
the coding style was not applied to this file ?

src/Widgets/Others/Help.js
```diff
@@ -3,6 +3,8 @@
 //Components
 import { ModuleView } from '../Components/ModuleView/ModuleView';
 
+import jQuery from 'jquery';
```
EricBoix, line 6: I was told that jQuery was only required for applications and not for packages.
When is it a good practice to use JQuery in a package ?

src/Widgets/Widgets.md
```diff
@@ -0,0 +1,102 @@
+### Camera Controller
```
EricBoix, line 1: This code seems to be imported from [UD-Viz-Template](https://github.com/VCityTeam/UD-Viz-Template/blob/master/Readme.md#camera-controller) (because eventually it belongs in UD-Viz).
Provide a PR in UD-Viz-Template that cleans up the source from the content that was moved here.

EricBoix (changes requested):

src/Widgets/Widgets.md
EricBoix: Already fixed by [this PR](https://github.com/VCityTeam/UD-Viz-Template/pull/34)

src/Templates/AllWidget/AllWidget.js
valentinMachado: yes this change is due to prettier formatter plugin.

EricBoix: I meant all the changes of this file seem to be coding style changes. That goes into the right direction.
Although what we want to avoid is going round in circles because different contributors have different coding styles (drawn from their respective editor). Hence my question: is _the_ coding of this code (as specified by the eslint configuration) now hardwired for this file ? For other files ? (this doesn't seem to be the case according to 
[the current version of .eslintrc.js](https://github.com/VCityTeam/UD-Viz/blob/master/.eslintrc.js))

In other terms shouldn't eslint be run in the CI to avoid such "noisy" side effects ?

src/Widgets/Others/Help.js
EricBoix: Lesson learned: JQuery usage indicates the dependency towards some server. Thus code making such usage could be placed in the Extensions sub-directory in order to indicate this dependency.

EricBoix (changes requested):
I did reopen a conversation on the coding style enforcement.

src/Templates/AllWidget/AllWidget.js
valentinMachado: eslint is okay with the both version (with parenthesis and without) meaning that running the eslint CI will not throw error but prettier put parenthesis. Iam not sure @clementcolin who writed this code originally is using a formatter. If this is the case it's not a formatter conflict.

This highlight the fact that eslint and prettier are not completly synchronise see this issue https://github.com/VCityTeam/UD-Viz/issues/181

EricBoix (approved):

src/Templates/AllWidget/AllWidget.js
clementcolin: I do use a formatter, hence it should be a "prettier" problem as eslint does not return any warning. Maybe there is a way to add a prettier to eslint ?

PR #242, opened 2021-10-08
Sparql module

DiegoVinasco:
Add a widget extension **SPARQL** to allow basic communication and interaction with RDF-Store data
- Module allows quering to RDF-Stores with SPARQL
- Query results can be displayed as a graph
- Visualized graph data can be used to select city objects in the UD-Viz view


package.json
```diff
@@ -25,6 +25,7 @@
   "homepage": "https://github.com/VCityTeam/UD-Viz/",
   "peerDependencies": {},
   "dependencies": {
+    "d3": "^7.0.1",
```
EricBoix, line 28: In order to keep track of package dependencies (e.g. only SPARQL uses d3, which can always be re-inferred by seaching "import d3" in all sources) with a comment in this file, maybe we could use this [json trick for npm](https://stackoverflow.com/a/45815391/5649243) ?

src/Widgets/Components/LayerManager/LayerManager.js
```diff
@@ -146,6 +146,60 @@ export class LayerManager {
     return undefined;
   }
 
+  /**
+   * Returns the city object which corresponds to a key,value pair in a tilesManager
+   * batch table. The first city object whos batch table entry matches the criteria
```
EricBoix, line 151: whos ---> whose

src/Widgets/Documents/README.md
```diff
@@ -11,7 +11,7 @@ The module is also extensible, which means that other modules can serve as plug-
 
 ## Basic functionnalities
 
-In the [demo](http://rict2.liris.cnrs.fr/UD-Viz/UD-Viz-Core/src/Modules/Documents/Example/Demo.html), you can try out the basic functionnalities of the document modules. It can:
+In the [demo](http://rict.liris.cnrs.fr/UDVDemo/UDV/UDV-Core/examples/DemoFull/Demo.html), you can try out the basic functionalities of the document modules. It can:
```
EricBoix, line 14: I would rather point to something local to this repository (like 
`UD-Viz/examples/DocumentWidget`) rather than to an installed demo that might be moved to another deployment server thus breaking this URL link.

src/Widgets/Extensions/SPARQL/Model/SparqlEndpointService.js
```diff
@@ -0,0 +1,54 @@
+import { RequestService } from '../../../../Components/Request/RequestService';
+
+/**
+ * The SPARQL Endpoint Service which contains connection information for
+ * a specific SPARQL Endpoint.
+ */
+export class SparqlEndpointService extends RequestService {
+  /**
+   * Creates a SPARQL Endpoint Service
```
EricBoix, line 9: This class does not create a SPARQL endpoint but it encapsulates the information designating a SPARQL Endpoint (for later consumption i.e. submitting requests/searches).
Maybe (?) a clearer comment would here be of the form: 
extends the given configuration with all the information required for accessing the designated SPARQL endpoint.

valentinMachado (approved):

PR #253, opened 2021-11-02
v2.34.1 Add SPARQL widget example

DiegoVinasco:


valentinMachado:
I can't launch the example since there is no config for the sparql widget. Could you provide me one ? @DiegoVinasco

DiegoVinasco:
Sure thing, sorry about that. I've added an example config with an icon for the widget. Let me know if you have any issues launching it

PR #275, opened 2022-01-05
Display a default style per mesh

LorenzoMarnat:
This PR allows to display a default style per GlTF mesh, so we can have different styles in a same tile. Those default styles are loaded directly from the 3DTiles. UD-Viz no longer overrides (by default) the GlTF materials. If a color is added in the json config file for a 3DTiles layer, this color will be used as default style for the whole layer.

The PR also allows to handle tiles with more than one mesh.


src/Components/3DTiles/3DTilesUtils.js
```diff
@@ -526,19 +535,20 @@ export function getMeshFromTile(tile) {
   }
 
   //Find the 'Mesh' part of the tile
-  while (!!tile.children[0] && !(tile.type === 'Mesh')) {
+  while (!!tile.children[0] && !(tile.children[0].type === 'Mesh')) {
```
mathieuLivebardon, line 538: why double exclamation mark ?

src/Components/3DTiles/3DTilesUtils.js
```diff
@@ -101,13 +101,18 @@ export function getVisibleTileCount(layer) {
...
- * @param {*} tileset The 3DTiles tileset.
+ * @param {*} rootTile The root tile of the 3DTiles tileset.
  * @param {*} tileId The tile id.
  */
-export function getTileInTileset(tileset, tileId) {
-  let tile = tileset.children.find((tile) => {
-    return tile.tileId === tileId;
-  });
+export function getTileInTileset(rootTile, tileId) {
+  if (rootTile.tileId === undefined) return undefined;
```
mathieuLivebardon, line 108: Maybe use this format of if :
if(!rootTile.tileId) return null;

src/Components/3DTiles/TilesManager.js
```diff
@@ -227,6 +251,21 @@ export class TilesManager extends EventSender {
...
+   * Sets the style of a particular tile.
+   * 
+   * @param {Int} tileId The tile
+   * identifier.
+   */
+  setDefaultStyleToTile(tileId) {
+    if(this.tiles[tileId]){
+      for(let i in this.tiles[tileId].cityObjects){
+        let meshId = this.tiles[tileId].cityObjects[i].meshId;
+        this.setStyle(this.tiles[tileId].cityObjects[i].cityObjectId, 'default' + tileId + 'm' + meshId);
```
clementcolin, line 264: Create a method to construct the style id from a tileId and meshId instead of hard coding it may be better

src/Components/3DTiles/3DTilesUtils.js
LorenzoMarnat: "Converts Object to boolean. If it was falsey (e.g. 0, null, undefined, etc.), it will be false, otherwise, true." ([!! operator](https://stackoverflow.com/questions/784929/what-is-the-not-not-operator-in-javascript))

mathieuLivebardon (approved):

PR #276, opened 2022-01-06
Jsdoc

clementcolin:
The first documentation can be found in the [jsdoc](https://github.com/VCityTeam/UD-Viz/tree/jsdoc) branch.

It is produced using [jsdoc](https://jsdoc.app/)

What misses before accepting this PR :
- [x] assert that the process to create the documentation is working fine
- [x] move sources to Documentation
- [x] document the publishing stage (after `npm run doc` one must commit and....)
- [x] For this PR remove the generated doc. Publishing will be done later on.


Readme.md
```diff
@@ -41,6 +41,22 @@ node index.js PATH_TO_UD-Viz 8000
...
+To generate the documentation : 
+
+```bash
+git clone https://github.com/VCityTeam/UD-Viz.git
+cd UD-Viz
+npm install
+npm run docs
+```
+
+
```
EricBoix, line 59: Ok the mardown sources seem to have been moved into docs/static/Doc/...
What is this static distinction for ? Is there a dynamic part ? Or is the distinction between source = static and target ? Or maybe you are using static for mardown type of source ?!
Independently from the static part, do you need the "Doc" repetition (this is already in docs, right?).

docs/static/readme.md
```diff
@@ -0,0 +1 @@
+This is the ud-viz documentation
```
EricBoix, line 1: Why is docs/static/Installation.md an empty file ? If it's a dummy for e.g. jsdoc can't this be stated in that file ?

package-lock.json
```diff
@@ -67,6 +67,12 @@
         }
       }
     },
+    "@babel/parser": {
```
EricBoix, line 70: Commiting the package-lock is a different task, isn't it ?
Is this really a requirement for this PR ?

src/Templates/AllWidget/AllWidget.js
```diff
@@ -401,26 +401,28 @@ export class AllWidget {
...
-      {
-        useColorTextureElevation: true,
-        colorTextureElevationMinZ: 144,
-        colorTextureElevationMaxZ: 622,
-        source: wmsElevationSource,
-      }
-    );
-    this.view.addLayer(wmsElevationLayer);
-    this.update3DView();
+    if(this.config['elevation_layer']){
```
EricBoix, line 404: Does this concern the jdoc process ? When not, please make it a separate PR.

EricBoix (changes requested):

src/Templates/AllWidget/AllWidget.js
clementcolin: I don't understand why this changer appears is in this PR. It should already be in UD-Viz. I'll try and erase it.

EricBoix (approved):

PR #297, opened 2022-02-17
Fix layer choice widget

clementcolin:
An old refacto of Tile.getMesh() to Tile.GetMeshes() was not finished, thus implying some bug. Fixed it in this PR

### DIFF
--- a/docs/html/Widgets_LayerChoice_views_LayerChoice.js.html
+++ b/docs/html/Widgets_LayerChoice_views_LayerChoice.js.html
@@ -260,7 +260,7 @@ export class LayerChoice extends Window {
         for(let j = 0; j &lt; tiles.length ; j++) {
           if (tiles[j].asAttributeInBatchTable('classe')) {
             let classe = tiles[j].batchTable.content.classe[tiles[j].cityObjects[0].batchId];  
-            htmlTiles += `&lt;p>&lt;input type="checkbox" id="checkbox_${i}_${j}" ${tiles[j].getMeshes().visible ? 'checked' : ''}>${classe}&lt;/input>&lt;/p>`;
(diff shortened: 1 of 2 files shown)


EricBoix (approved):

docs/html/Widgets_LayerChoice_views_LayerChoice.js.html
```diff
@@ -260,7 +260,7 @@ <h1 id='page-title' class="page-title">
         for(let j = 0; j &lt; tiles.length ; j++) {
           if (tiles[j].asAttributeInBatchTable('classe')) {
             let classe = tiles[j].batchTable.content.classe[tiles[j].cityObjects[0].batchId];  
-            htmlTiles += `&lt;p>&lt;input type="checkbox" id="checkbox_${i}_${j}" ${tiles[j].getMesh().visible ? 'checked' : ''}>${classe}&lt;/input>&lt;/p>`;
+            htmlTiles += `&lt;p>&lt;input type="checkbox" id="checkbox_${i}_${j}" ${tiles[j].getMeshes().visible ? 'checked' : ''}>${classe}&lt;/input>&lt;/p>`;
```
mathieuLivebardon, line 263: Why this change?
Also I have the impression that the doc is not up to date in all ways.

mathieuLivebardon (approved):
For me it's good.
Think to take a look at the focus button in the LayerChoice widget.

docs/html/Widgets_LayerChoice_views_LayerChoice.js.html
clementcolin: It was due to an old refacto, it should not be in this pr, you're right.

Concerning doc, maybe we should reproduce it after a PR ? Or at each UD-Viz published version ?

mathieuLivebardon: We need to set up a workflow about it. I think updating the doc at each release is a good idea.

PR #310, opened 2022-03-10
Post it

valentinMachado:


src/Views/AssetsManager/AssetsManager.js
```diff
@@ -10,7 +10,9 @@ const THREEUtils = require('../../Game/Components/THREEUtils');
 /**
  * Default material used by native objects
  */
-const DEFAULT_MATERIAL = new THREE.MeshLambertMaterial({ color: 0x00ff00 });
+const DEFAULT_MATERIAL = new THREE.MeshLambertMaterial({
```
mathieuLivebardon, line 13: Why it's a global variable ?

mathieuLivebardon (commented):
- The PR is missing a description (we need it in order to understand the title)
- Is this line of DistantGame good ? [#L11](https://github.com/VCityTeam/UD-Viz/blob/457d3bc067867c26d9c5f6aa117364fc94b54c3f/src/Templates/Game/DistantGame/DistantGame.js#L11) => VMA: yes it is euh no you are right fix with this one https://github.com/VCityTeam/UD-Viz/pull/310/commits/e2502ad96d534e1ba226af6a1b971ddce9b39c96

src/Views/AssetsManager/AssetsManager.js
valentinMachado: to avoid to write "this" before it and because it's not going to change (const) but it could be a member attribute since it's only use in this file

mathieuLivebardon:
All Game example doesn't work
Console when launch ./examples/LocalGame.html
2022-03-10_14h12_07

valentinMachado:
forgot to run npm run debug, should be fixed by the last commit

Issue #325, opened 2022-03-29
CityObjects does not refresh displayed attributes on second selection   


EricBoix:
There appears to be different bugs

## How to reproduce the first bug -- _FIXED_
1. Launch UD-Viz examples (e.g. `npm run build` and `python3 -m http.server` and web-browse on the provided URL...)
2. Select the AllWdigets example of the mosaic
3. Toggle the `CityObjects` widget on
4. Select a first building in the scene: the selected building gets properly highlighted in the scene and the attributes of that building get properly displayed in the CityObject sub-window (ok)
5. Hit the `clear-selection` button: the previously selected building gets properly UN-highlighted (ok) BUT the attributes of that previously selected building remain displayed in he CityObject sub-window (bug)

## How to reproduce the second bug -- _FIXED_

1. Launch UD-Viz examples (e.g. `npm run build` and `python3 -m http.server` and web-browse on the provided URL...)
2. Select the AllWdigets example of the mosaic
3. Toggle the `CityObjects` widget on
4. Select a first building in the scene: the selected building gets properly highlighted and attributes of that building get properly displayed (ok)
5. Select a second building in the scene:  the newly selected building gets properly highlighted BUT its attributes are not displayed (the previously displayed attributes of the first building keep being displayed).

Notice that, if prior to step 4 you first hit the `clear-selection` button (and although the attributes remain erroneously displayed, refer to first above bug) THEN when realizing step 5 (second selection) the attributes of the newly selected building get properly refreshed.

## How to reproduce the third bug -- _FIXED_

1. Launch UD-Viz examples (e.g. `npm run build` and `python3 -m http.server` and web-browse on the provided URL...)
2. Select the AllWdigets example of the mosaic
3. Toggle the `CityObjects` widget on
4. Drag the created sub-window upwards : the sub-window bottom line never appears, that is the sub-window is vertical length is not properly set (bug)

Note if you open the filter second sub-window (hit the `select` button of the Filter section of the first sub-window) and
drag that `City Objects - Filters` sub-window upwards then the same defect appears (that second sub-window is vertical length is not properly set)

## How to reproduce the fourth bug

1. Launch UD-Viz examples (e.g. `npm run build` and `python3 -m http.server` and web-browse on the provided URL...)
2. Select the AllWdigets example of the mosaic
3. Toggle the `CityObjects` widget on : a `City Object` sub-window appears (ok)
[...]

PR #355, opened 2022-05-19
Fix Various bugs 

clementcolin:
Following [this issue](https://github.com/VCityTeam/UD-Viz/issues/325), I fixed various bugs :

- The default style is now correctly set to new 3DTiles layer, and thus does not generate error when selecting an object
- The Event 'CityObjectChanged' or 'CityObjectSelected' is send when needed, allowing to refresh displayed attributes
- The FocusCameraOn method now properly zoom on an object, using its position, instead of the camera position
- The picker does now avoid to pick any object in non-visible layer


src/Components/3DTiles/TilesManager.js
```diff
@@ -131,17 +130,24 @@ export class TilesManager extends EventSender {
...
   }
 
   focusCamera() {
-    let coordinates = this.view.camera.position();
-    let extent = this.layer.extent;
-    coordinates.x = (extent.east + extent.west) / 2; 
-    coordinates.y = (extent.north + extent.south) / 2;
-    focusCameraOn(this.view,this.view.controls,coordinates);
+    if(this.layer.isC3DTilesLayer){
+      let coordinates = this.view.camera.position();
```
mathieuLivebardon, line 140: Prefer `const` than `let` when it's possible

LorenzoMarnat:
An error occurs if a 3D Tiles layer has no color 

```json
  "3DTilesLayers": [
    {
      "id": "3d-tiles-layer-building",
      "url": "https://dataset-dl.liris.cnrs.fr/three-d-tiles-lyon-metropolis/Demo/UD-Demo-vcity-py3dtilers-lyon/Lyon-1_2018_Textured/tileset.json"
    },
```
image

Not having a color field in a 3D Tiles layer allows to use the material from the glTF (adding a color will overwrite the materials, destroying textures, colors, etc)

LorenzoMarnat:
When using CityOBject or 3DTiles Debug widgets with LODs, the style is applied to the first intersected feature, even if the feature is in an invisible tile (this bug was existing before this PR).

bug_select

LorenzoMarnat (changes requested):
[This bug](https://github.com/VCityTeam/UD-Viz/pull/355#issuecomment-1132673435) should be resolved before merging the PR.

[The bug on invisible tiles](https://github.com/VCityTeam/UD-Viz/pull/355#issuecomment-1132681628) could be resolved later, since the bug was already existing before

src/Components/3DTiles/3DTilesUtils.js
```diff
@@ -42,14 +42,16 @@ export function getFirstTileIntersection(intersects) {
...
-    if (!!geomAttributes && !!geomAttributes._BATCHID) {
-      if (!first_inter) {
-        first_inter = inter;
-        dist_min = inter.distance;
-      } else if (inter.distance < dist_min) {
-        first_inter = inter;
-        dist_min = inter.distance;
+    if(inter.object.visible){
+      let geomAttributes = inter.object.geometry.attributes;
+      if (!!geomAttributes && !!geomAttributes._BATCHID) {
```
mathieuLivebardon, line 47: Why double `!` ?

src/Components/Camera/CameraUtils.js
```diff
@@ -25,14 +25,17 @@ export function focusCameraOn(view, controls, targetPos, options = {}) {
       const duration = options.duration || null;
       const verticalDist = options.verticalDistance || 800;
       const horizontalDist = options.horizontalDistance || 1000;
-
+
```
mathieuLivebardon, line 28: Useless changes

src/Widgets/CityObjects/CityObjectModule.js
```diff
@@ -34,7 +34,6 @@ export class CityObjectModule {
      */
     this.provider = new CityObjectProvider(layerManager);
     this.provider.setSelectionStyle(config.cityObjects.styles.selection);
-
```
mathieuLivebardon, line 37: Same

mathieuLivebardon:
I will create an issue about coding style. I think we should avoid formatting war.

src/Components/3DTiles/3DTilesUtils.js
mathieuLivebardon: Ok I didn't know, then it doesn't matter, it will disappear with the PR of formatting style.

LorenzoMarnat (approved):

mathieuLivebardon:
Rebase and then is ok

mathieuLivebardon (approved):

PR #384, opened 2022-07-12
Navigation Widget

CorentinGaut:
## Navigation UI
Creation of an UI to navigated differenty in the UD-Viz scene and have more information about your localisation.

The UI is composed of : 
- A slider zoom that move the camera on his point of view with a specific constant.
- A compass to indicate north in the view

This UI work on the [PlanarView](http://www.itowns-project.org/itowns/docs/#api/View/PlanarView) of itowns which is different to the [Navigation Widget](http://www.itowns-project.org/itowns/docs/#api/Widgets/Navigation) of this library that only work on [GlobeView](http://www.itowns-project.org/itowns/docs/#api/View/GlobeView).

An example of the widget is integrate in the AllWidget page.


src/Widgets/Extensions/Navigation/Navigation.js
```diff
@@ -0,0 +1,69 @@
...
+import { ModuleView } from '../../Components/ModuleView/ModuleView';
+import * as THREE from 'three';
+import './navigation.css';
+
+export class Navigation extends ModuleView {
+  constructor(view3D) {
+    super();
+    
+    //Scene renderer
+    let renderer = view3D.mainLoop.gfxEngine.renderer;
```
mathieuLivebardon, line 11: prefer const than let when it's possible

src/Widgets/Extensions/Navigation/navigation.css
```diff
@@ -0,0 +1,106 @@
...
+    /* Magic trick to center */
+    left: 0;
+    right: 0;
+    margin-left: auto;
+    margin-right: auto;
+    bottom: 70px;
+    z-index: 10;
+    background-color: #00000088;
+    list-style-type: none;
+
```
mathieuLivebardon, line 15: remove this space

clementcolin:
Dumb question from the outside, but why do not we extend the itowns widget that works only on globe view to make it work on planar view? It is meant to be integrated in the futur ?

CorentinGaut:
@clementcolin it's a different approach to Itowns zoom but in the futur it will be an enhancement of the navigation widgets from Itowns

PR #408, opened 2022-07-26
fix point cloud size parameter

clementcolin:


valentinMachado:
this fix look to have been included here https://github.com/VCityTeam/UD-Viz/blob/8ac70f75ec3784e97ded4b0b96952c89d422a822/src/Components/ItownsViewUtils.js#L61 
is this pr still relevant ?

PR #459, opened 2022-11-25
Support 'x-bil;bits=32' texture format for Elevation layer

LorenzoMarnat:
`useColorTextureElevation`,  `colorTextureElevationMinZ` and `colorTextureElevationMaxZ` are now mandatory only if the elevation format is 'png' or 'jpeg'. This allows to support elevation formats such as 'x-bil;bits=32'


mathieuLivebardon (approved):
⛸️

PR #460, opened 2022-11-25
Use 'crs' instead of deprecated 'projection' in layers

LorenzoMarnat:
Fix the following warning: 

image


mathieuLivebardon (commented):
Should we change here too ?
https://github.com/VCityTeam/UD-Viz/blob/a2263e95c728e65392e8548585f7942b4763cd2f/src/Widgets/BaseMap/BaseMapWindow.js#L50

mathieuLivebardon (approved):
↔️

PR #503, opened 2023-01-24
Added dynamic scripts src inside html

LinkJerome:
- Added string-replace-middleware
- Added RUN_MODE template key
- Dynamic script sources

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -21,6 +21,7 @@
     "puppeteer": "^19.4.1"
   },
   "devDependencies": {
-    "webpack-node-externals": "^2.5.2"
+    "webpack-node-externals": "^2.5.2",
+    "string-replace-middleware": "^1.0.2"
   }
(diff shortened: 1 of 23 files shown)


packages/node/package.json
```diff
@@ -21,6 +21,7 @@
     "puppeteer": "^19.4.1"
   },
   "devDependencies": {
-    "webpack-node-externals": "^2.5.2"
+    "webpack-node-externals": "^2.5.2",
+    "string-replace-middleware": "^1.0.2"
```
valentinMachado, line 25: string replace middleware should be a dependencie

valentinMachado (changes requested):

packages/node/package.json
LinkJerome: You're right, I have fixed it

valentinMachado (approved):

PR #520, opened 2023-02-09
Fix temporal example

DiegoVinasco:
just needed to add the batch table extension in the config 😉


packages/browser/examples/TemporalWidget.html
```diff
@@ -80,11 +80,11 @@
...
 
         // TEMPORAL MODULE TODO_ISSUE broken
-        // const temporalModule = new udvizBrowser.Widget.TemporalModule(
-        //   app.getFrame3DPlanar().getLayerManager().tilesManagers[0],
-        //   configs['temporal']
-        // );
-        // app.addWidgetView('temporal', temporalModule.view);
+        const temporalModule = new udvizBrowser.Widget.TemporalModule(
+          app.getFrame3DPlanar().getLayerManager().tilesManagers[0],
+          configs['temporal']
```
valentinMachado: this config is not loaded, don't pass anything

valentinMachado (approved):

packages/browser/examples/TemporalWidget.html
valentinMachado: I just did it

PR #524, opened 2023-02-10
Delete deprecated or unused 3DTiles methods

LorenzoMarnat:
A lot of methods in `3DTilesBuildingUtils.js` or `3DTilesUtils.js` seems to be either deprecated or never used. I think those methods aren't used by anyone or by any demo, they should be deleted


valentinMachado:
Great !
How have you determine that they were deprecated ?

LorenzoMarnat:
It seems most of those methods were created before `TilesManager`, `StyleManager`, `Tile` and `CityObject` classes. Those classes allow us to do more easily what the deleted methods do

And since the deleted methods aren't either referenced in UD-Viz or exposed in the API, I guess we can delete the code without creating issues

valentinMachado:
I was wondering how to know if an external project was using these methods but since they are not exposed in the API you right it should be fine !

valentinMachado (approved):

Issue #487, opened 2023-01-11
Crs of geoJSON layer should be in config ?


valentinMachado:
https://github.com/VCityTeam/UD-Viz/blob/b6ff65f27e77dd8b82c73b1ab33e0244c9885269/packages/browser/src/Component/Itowns/AddLayerFromConfig.js#L164

Here a crs projection is passed to the source layer of geojson and is defined in config of these layers.
Since planarview has been created with an extent with its own crs and other layer as well (extent is passed at the creation)
This should be the case of geojsonlayers as well right ? => remove crs attribute from config and pass crs extent at creation

PR #526, opened 2023-02-10
Add iTowns LabelLayers

LorenzoMarnat:
Method to set up and add LabelLayers in the itowns view.

The source of a LabelLayer can be either a WFS source or a File source.

The features in the source must be Point geometries.

Add an example in AllWidget example

### DIFF
--- a/packages/browser/examples/AllWidget.html
+++ b/packages/browser/examples/AllWidget.html
@@ -79,7 +79,8 @@
 
         udvizBrowser.addLabelLayers(
           configs['labels'],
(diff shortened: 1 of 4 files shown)


valentinMachado:
The label layer config take a crs, is this one could be different of the extent one ?
Related to https://github.com/VCityTeam/UD-Viz/issues/487

LorenzoMarnat:
> The label layer config take a crs, is this one could be different of the extent one ?

The CRS should always be the same as the extent one

valentinMachado (approved):

PR #530, opened 2023-02-17
fix cylinder pin rotation + clean

valentinMachado:
fix pin's cylinder rotation + remove unused code in geocoding


DiegoVinasco:
the fix is perfect
image

however, I must admit that I removed some of the symbolic links in the dataset server which broke a demo.

Could we pull [PR 592](https://github.com/VCityTeam/UD-Viz/pull/529) first?

DiegoVinasco (approved):

valentinMachado:
> the fix is perfect image
> 
> however, I must admit that I removed some of the symbolic links in the dataset server which broke an example. (I didn't realize this would break the puppeteer test)
> 
> Could we pull [PR 592](https://github.com/VCityTeam/UD-Viz/pull/529) first?

Yeah I realized that as well, I have done it

Issue #531, opened 2023-02-17
reduce npm run validate-links threshold to 9


valentinMachado:
see https://github.com/VCityTeam/UD-Viz/pull/532

valentinMachado:
Could nice to threat this one as well https://github.com/VCityTeam/UD-Viz/issues/516 since they are very coupled

Issue #517, opened 2023-02-02
Last travis job remark validate link cant be launch in local


valentinMachado:
Add remark-cli and remark-validate-links in dev dependencies of the ud-viz-mono-repo package.json.
So all travis job can be run automatically in local.

PR #532, opened 2023-02-17
Last travis job remark validate link cant be launch in local #517

valentinMachado:
The remark validate-links can now be runned with npm run validate-links CLI

The travis job was broken and I realized there were 56 warnings instead of 9 which is the desired threshold
I began to fix some of them but some are out of my knowledge scope.

So I put the threshold at 50 so test can pass  for now and I opened an issue https://github.com/VCityTeam/UD-Viz/issues/531
To participate assign yourself in the issue then we could fix a point together to resolve it


valentinMachado:
I change the threshold to 74 since there is more dead links on master now. see https://github.com/VCityTeam/UD-Viz/issues/531

mathieuLivebardon (approved):
OK

PR #540, opened 2023-03-01
add exports

clementcolin:


packages/browser/src/Component/Itowns/AddLayerFromConfig.js
```diff
@@ -20,7 +20,7 @@ const $3DTemporalTileset = Widget.$3DTemporalTileset;
  * @param {itowns.View} itownsView - the itowns view
  * @returns {itowns.C3DTilesLayer} A 3D Tiles Layer
  */
-function setup3DTilesLayer(layer, layerManager, itownsView) {
+export function setup3DTilesLayer(layer, layerManager, itownsView) {
```
valentinMachado, line 23: there is a case where you have to use this function instead of add3DTilesLayers ?

packages/browser/src/Component/Itowns/Itowns.js
```diff
@@ -11,6 +11,12 @@ export { LayerManager } from './LayerManager/LayerManager';
 /** @class TilesManager */
 export { TilesManager } from './3DTiles/TilesManager';
 
+/**
+ * @class CityObject
+ * @class CityObjectID
+ */
+export { CityObject, CityObjectID } from './3DTiles/Model/CityObject';
```
valentinMachado, line 18: why expose these classes here ? since Itowns.js is not expose in the API they are no going to be expose in @ud-viz/browser API

packages/browser/src/Component/Itowns/AddLayerFromConfig.js
clementcolin: Yes, I want to be able to add a 3DTiles layer without a config file / dynamically

valentinMachado: :+1:

packages/browser/src/Component/Itowns/Itowns.js
valentinMachado: Yes the index.js is the right place, you can do it the way you did it but it's better practice to expose it in the index.js (@ud-viz/browser API) so it make simpler to check what can be use outside the package (for refacto purpose, or to detect unused part of the code). Furthermore if you were using a bundle import like in the example it would be not possible to access Itowns.

Issue #534, opened 2023-02-21
Adapt ud-viz API to work with UD-IMUV


valentinMachado:
Implement changes required for UD-IMUV

valentinMachado:
external context when reset should remove its object3D from parent if there were already a frame3D

PR #542, opened 2023-03-01
Adapt ud viz api to work with ud imuv #534

valentinMachado:


valentinMachado:
defaultGameService should not be launch in Test by default => test cant be launch elsewhere than ud-viz
one solution should be to remove gamesocketservice from express app wrapper and to force the user to build thread bundle then to launch a gamesocket service

mathieuLivebardon (approved):
🧴

PR #545, opened 2023-03-03
fix by reading client instead of offset

valentinMachado:


mathieuLivebardon (approved):
🖱️🔥🦊

mathieuLivebardon (changes requested):
Example doesn't work in production mode 
image

- `npm run start`
- Open `http://localhost:8000/packages/browser/examples/MultiPlayerGamePlanar.html`

valentinMachado:
> Example doesn't work in production mode image
> 
> * `npm run start`
> * Open `http://localhost:8000/packages/browser/examples/MultiPlayerGamePlanar.html`

Yeah the host script was not launching the game socket service, but this is going to be fixed by https://github.com/VCityTeam/UD-Viz/pull/542 that should be reviewed in priority

mathieuLivebardon (approved):
👌

Issue #516, opened 2023-02-02
Doc : static doc file in src


mathieuLivebardon:
Since static files are meant to be docs/static/ folder, ones in sources should be moved in this folder.

PR #548, opened 2023-03-03
Fix broken links

LorenzoMarnat:
The goal of this PR is to fix the links in UD-Viz documentation

`remark -u validate-links .` still returns 30 warnings (instead of 74)


valentinMachado:
@LorenzoMarnat Iam going to solve #516 on your branch

DiegoVinasco:
This is great. In addition to this work however, we might want to consider a flaw with remark validate links...

Bascially, Remark doesn't validate absolute links (which are accepted by github in markdown documents). It only checks relative links or URLs.

This means that we might still have some broken links. This was worked around in the private repository [here](https://github.com/VCityTeam/VCity/blob/master/.travis.yml) by using a `sed` command with some pipes to prefix absolute markdown links with a given URL:
```bash
export BASE_URI='https://github.com/VCityTeam/VCity/blob/master'
files=$(find . -name '*.md')
for f in $files
do
  grep -E '\[{1}.*\]{1}\({1}/{1}.*\){1}' "$f" | sed -i "s|](/|]($BASE_URI/|g" "$f"
done
remark -u validate-links .
```

I think it would be a good idea to do the same in UD-Viz. Maybe in a different PR? Any thoughts?

valentinMachado:
> This is great. In addition to this work however, we might want to consider a flaw with remark validate links...
> 
> Bascially, Remark doesn't validate absolute links (which are accepted by github in markdown documents). It only checks relative links or URLs.
> 
> This means that we might still have some broken links. This was worked around in the private repository [here](https://github.com/VCityTeam/VCity/blob/master/.travis.yml) by using a `sed` command with some pipes to prefix absolute markdown links with a given URL:
> 
> ```shell
> export BASE_URI='https://github.com/VCityTeam/VCity/blob/master'
> files=$(find . -name '*.md')
> for f in $files
> do
>   grep -E '\[{1}.*\]{1}\({1}/{1}.*\){1}' "$f" | sed -i "s|](/|]($BASE_URI/|g" "$f"
> done
> remark -u validate-links .
> ```
> 
> I think it would be a good idea to do the same in UD-Viz. Maybe in a different PR? Any thoughts?

Since this PR is called "Fix broken links" imo we should let this open. I'm not sure to understand why there could be absolute link in .md, could you give an example of link that remark doesnt detect ?

valentinMachado:
Actually absolute link should be consider as a bad practice since the root of the path will depend of the context in which it's interpreted (like vscode workspaces or github organization).

Iam going to add in the script validate-links a code to detect absolute links and throw an error if one is found.

docs/static/Devel/Developers.md
```diff
@@ -38,30 +38,31 @@ consists in
...
-| `npm run docs`           | Run `npm run docs-shared`, `npm run docs-browser`, `npm run docs-node`, and `npm run docs-home`                                                                                    |
-| `npm run dev-docs`       | Launch a watcher for generating and debugging the documentation. See [here](./docs/Readme.md) for more information                                                                 |
-| `npm run host`           | host the bundle with an [ExpressAppWrapper](./bin/host.js). <br>http://locahost:8000/                                                                                              |
-| `npm run start`          | Run `npm run build-browser` and `npm run host`                                                                                                                                     |
+| `npm run reset`          | Reinstalls npm dependencies. This script runs `npm run clean` and `npm install` command                                                                                                                                                                               |
```
EricBoix, line 45: Why is there so much noise with many lines that appear as removed and then replaced with apparently identical content (at least on github diff view) ?

docs/static/Devel/Configuring-the-Document-model-of-an-UDV-deployment.md
```diff
@@ -1,305 +0,0 @@
-# WARNING : Outdated document
```
EricBoix, line 1: Damned for once we had some structured documentation... :-/ Too bad its so updated that fixing it is worthless.

docs/static/Devel/LocalGameTutorial.md
```diff
@@ -1,828 +0,0 @@
-<!-- @format -->
-
-# Local Game tutorial :lion:
```
EricBoix, line 3: Again  that's quite some bit of documentation that is flushed out. There is really nothing to save/adapt/maintain ?

docs/static/Devel/UD_Viz_Browser/SPARQL/Readme.md
```diff
@@ -11,23 +11,23 @@ The basic functionalities of the SPARQL module include:
 
 ## Usage
 
-For an example of how to the SPARQL Widget to a UD-Viz web application see the [SPARQLWidget example](https://github.com/VCityTeam/UD-Viz/blob/master/packages/browser/examples/SPARQLWidget.html)
+For an example of how to the SPARQL Widget to a UD-Viz web application see the [SPARQLWidget example](https://github.com/VCityTeam/UD-Viz/blob/master/examples/SPARQLWidget.html)
```
EricBoix, line 14: The pointed markdown seems to live in this repo. Why is this not a relative link ?

packages/node/Readme.md
```diff
@@ -31,7 +31,7 @@ See [here](https://github.com/VCityTeam/UD-Viz/blob/master/Readme.md#getting-sta
 
 ## Developers
 
-For pre-requisites see [here](https://github.com/VCityTeam/UD-Viz/blob/master/Readme.md#pre-requisites).
+For pre-requisites see [here](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/Developers.md#pre-requisites).
```
EricBoix, line 34: Internal link ?

packages/browser/Readme.md
```diff
@@ -45,7 +45,7 @@ See [here](https://github.com/VCityTeam/UD-Viz/blob/master/Readme.md#getting-sta
 
 ## Developers
 
-For pre-requisites see [here](https://github.com/VCityTeam/UD-Viz/blob/master/Readme.md#pre-requisites).
+For pre-requisites see [here](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/Developers.md#pre-requisites).
```
EricBoix, line 48: Internal link ?

package.json
```diff
@@ -16,6 +16,7 @@
     "test": "cross-env NODE_ENV=production node ./bin/test.js",
     "assert-code": "npm run eslint && npm run test",
     "validate-links": "node ./bin/remarkValidateLinks.js",
+    "travis-jobs": "npm run assert-code && npm audit --audit-level=low && npm run validate-links",
```
EricBoix, line 19: Maybe neutralize the target name ? (name it `ci-jobs`, or  something similar, just for the day when we might move away from Travis-ci)

EricBoix (changes requested):

docs/static/Devel/Developers.md
valentinMachado: git diff introduce this and went wrong

docs/static/Devel/LocalGameTutorial.md
valentinMachado: DITTO it would be more simple to make a new one

docs/static/Devel/UD_Viz_Browser/SPARQL/Readme.md
valentinMachado: All .md in docs/static are going to generate an html page in the online documentation, if the link is relative it will be broken online. That's why I think we should prefer url. Note that all .md are now going to be consider by jsdoc we should then replace all links by URL ? What do you think ?

DiegoVinasco: What about the case when we add a link to a new document that only exists in a local commit or branch. Here a link towards the online jsdoc would be broken.

This might be an edge case, but I think we need to define how to resolve it.

valentinMachado: Right, the CI command will not pass since these links do not exist yet. To overcome this we could adjust FAILURE_THRESHOLD variable to temporary pass the CI command. But it's quite an hack since we should then reset everytime the FAILURE_THRESHOLD variable...

valentinMachado: Just a thought... what about before to launch remark validate links we replace prefix "https://github.com/VCityTeam/UD-Viz/blob/master/" by "http:localhost:8000/" and we launch a local server on the UD-Viz repo. Url should be fine then we undo the prefix modification ?

Still sound like a hack but it should works, what about that ?

valentinMachado: I have created an issue #562 to think about how to solve the online documention link problem.
Since this PR is solving all dead links in github/desktop environment we could close it (even if some url could be relative path, this problem belong to #562) @DiegoVinasco and @EricBoix do you agree ?

DiegoVinasco: Yeah this issue can easily fall outside the scope of the PR. Sounds good to me.

DiegoVinasco (approved):

package.json
DiegoVinasco: One last request, could the developer readme be updated to match the change?

DiegoVinasco (approved):

mathieuLivebardon (approved):

PR #576, opened 2023-03-21
style(assetmanager) - loading bar centered

valentinMachado:
you can comment [these lines](https://github.com/VCityTeam/UD-Viz/blob/f69d2203583853df93306ccd41880c9f457f4725/packages/browser/src/Component/AssetManager/AssetManager.js#L207) to actually see the loading view

### DIFF
--- a/packages/browser/src/Component/AssetManager/AssetManager.css
+++ b/packages/browser/src/Component/AssetManager/AssetManager.css
@@ -10,8 +10,6 @@
 }
 
 .barBackground-Assets {
-  width: 50%;
-  margin-left: 25%;
   margin-top: 5px;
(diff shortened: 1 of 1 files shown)


mathieuLivebardon (commented):
I think you can remove the CSS class `.assetsloadingView` which seems useless.

valentinMachado:
> I think you can remove the CSS class `.assetsloadingView` which seems useless.

the 3D label is not centered without this class

mathieuLivebardon (approved):
> > I think you can remove the CSS class `.assetsloadingView` which seems useless.
> 
> the 3D label is not centered without this class

Right, so is good for me.

PR #578, opened 2023-03-22
Outliers for 3DTiles

clementcolin:
Add 3DTilesUtils.js method to create outlier of 3DTiles geometry, which should gives a result similar to : 
outliers


packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,4 +1,145 @@
...
+ */
+export function createOutliersOfTileset(tilesManager) {
+  tilesManager.addEventListener(
+    TilesManager.EVENT_TILE_LOADED,
+    function (tile) {
+      if (tilesManager.tiles[tile.tileId] != undefined) {
+        if (tilesManager.tiles[tile.tileId].cityObjects != undefined) {
+          const geom =
+            tilesManager.tiles[tile.tileId].getObject3D().children[0]
+              .children[0];
```
LorenzoMarnat, line 19: Maybe use `const geom = tilesManager.tiles[tile.tileId].getMeshes()[0];` instead of `...getObject3D().children[0].children[0]`

Since `const geom = ...getMeshes()[0];` will apply the outline only to the first mesh, you could also do `const meshes = ...getMeshes();` then loop on meshes

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,4 +1,145 @@
...
+ */
+export function createOutliersWithBatchIDOfTileset(tilesManager) {
+  tilesManager.addEventListener(
+    TilesManager.EVENT_TILE_LOADED,
+    function (tile) {
+      if (tilesManager.tiles[tile.tileId] != undefined) {
+        if (tilesManager.tiles[tile.tileId].cityObjects != undefined) {
+          const geom =
+            tilesManager.tiles[tile.tileId].getObject3D().children[0]
+              .children[0];
```
LorenzoMarnat, line 68: Ditto

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,4 +1,145 @@
...
+import { TilesManager } from './TilesManager';
+
+/**
+ * Create a geometric outliers for each tile of a tileset when its geometry is loaded
+ * Each outlier tile geometry does not keep the batchid for each outliers.
+ * May be slow to create / load.
+ *
+ * @param {TilesManager} tilesManager A TilesManage object from UD-Viz.
+ */
+export function createOutliersOfTileset(tilesManager) {
```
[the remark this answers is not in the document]
valentinMachado: its more appendWireframeToTileset since you add the wireframe object directly in the tile mesh

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,4 +1,145 @@
...
+  tilesManager.addEventListener(
+    TilesManager.EVENT_TILE_LOADED,
+    function (tile) {
+      if (tilesManager.tiles[tile.tileId] != undefined) {
+        if (tilesManager.tiles[tile.tileId].cityObjects != undefined) {
+          const geom =
+            tilesManager.tiles[tile.tileId].getObject3D().children[0]
+              .children[0];
+          if (!geom.hasOutlier) {
```
valentinMachado, line 20: when creating a wireframe object, tag it in userData with something like isWireframe, then to check if an object has already a wireframe attach to it do something like:
let hasWireframe = false;
object.traverse((child)=>{
if(!child.isWireframe)return;
hasWireframe = true;
})

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,4 +1,145 @@
...
+              let i = 0;
+              i < geom.geometry.attributes.position.count;
+              i += 3
+            ) {
+              indices.push(i, i + 1, i + 2);
+            }
+            geom.geometry.setIndex(indices);
+
+            // Create Outliers
+            const geo = new THREE.EdgesGeometry(geom.geometry, 30); // or WireframeGeometry
```
valentinMachado, line 40: Apparently there is a WireframeGeometry
[ref](https://discourse.threejs.org/t/proper-way-of-adding-and-removing-a-wireframe/4600)

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,4 +1,145 @@
...
+export function createOutliersWithBatchIDOfTileset(tilesManager) {
+  tilesManager.addEventListener(
+    TilesManager.EVENT_TILE_LOADED,
+    function (tile) {
+      if (tilesManager.tiles[tile.tileId] != undefined) {
+        if (tilesManager.tiles[tile.tileId].cityObjects != undefined) {
+          const geom =
+            tilesManager.tiles[tile.tileId].getObject3D().children[0]
+              .children[0];
+          if (!geom.hasOutlier) {
```
valentinMachado, line 69: ditto

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
clementcolin: WireframeGeometry will show each triangle of a mesh.
Here, I wanted to only make the edges of the mesh, not all triangles. We can also chose what define an edge : 
thresholdAngle — An edge is only rendered if the angle (in degrees) between the face normals of the adjoining faces exceeds this value. default = 1 degree.
(ref)[https://threejs.org/docs/#api/en/geometries/EdgesGeometry]

valentinMachado: ok got it, could be nice to make the thresholdAngle a parameter of the function with a default value at 30

examples/3DTilesOutliers.html
```diff
@@ -0,0 +1,67 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>3D Tiles Outliers example</title>
```
valentinMachado, line 4: Rename example file 3DTilesWireframe + html title and add a mosaic image in index.html

examples/3DTilesOutliers.html
```diff
@@ -0,0 +1,67 @@
...
+  </head>
+  <body>
+    <script src="../packages/browser/dist/RUN_MODE/bundle.js"></script>
+
+    <script type="text/javascript">
+      udvizBrowser.FileUtil.loadMultipleJSON([
+        './assets/config/extent_lyon.json',
+        './assets/config/frame3D_planars.json',
+        './assets/config/layer/3DTiles.json',
+        './assets/config/layer/base_maps.json',
```
valentinMachado, line 17: maybe it's not necessary to load and add the base map since the example only show a 3DTiles feature ? 
If you think that base map should be loaded then I think elevation should be as well.

examples/3DTilesOutliers.html
```diff
@@ -0,0 +1,67 @@
...
+        // ADD LAYERS
+        udvizBrowser.add3DTilesLayers(
+          configs['3DTiles'],
+          frame3DPlanar.layerManager,
+          frame3DPlanar.itownsView
+        );
+
+        // ADD OUTLIERS
+        for(const tilesManager of frame3DPlanar.layerManager.tilesManagers)
+          tilesManager.addEventListener(udvizBrowser.TilesManager.EVENT_TILE_LOADED,function(tile){udvizBrowser.appendWireframeToTileset(tile);});
```
valentinMachado, line 63: I modified this line

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,5 +1,133 @@
 import * as THREE from 'three';
 
+/**
+ * Create a geometric outliers for each tile of a tileset when its geometry is loaded
```
valentinMachado, line 4: Update description with append (and not create) and wireframe (and not outliers) could be something like:
Traverse a THREE.Object3D and append in children for each Object3D with a geometry a THREE.LineSegment representing its wireframe

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,5 +1,133 @@
...
+ * Each outlier tile geometry does not keep the batchid for each outliers.
+ * May be slow to create / load.
+ *
+ * @param {THREE.Object3D} tile  A 3DObject (should be a tile send by Tilesmanager.EVENT_TILE_LOADED)
+ * @param {number} threshOldAngle  An edge is only rendered if the angle (in degrees) between the face normals of the adjoining faces exceeds this value. default = 1 degree.
+ */
+export function appendWireframeToTileset(tile, threshOldAngle = 30) {
+  if (
```
valentinMachado, line 12: remove these if and remove the for loop to write something like:

tile.traverse((child)=>{
if(child.geometry && child.geometry.isBufferGeometry){
//do stuff
})

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,5 +1,133 @@
...
+        });
+        const edges = new THREE.LineSegments(geo, mat);
+        geom.add(edges);
+      }
+    }
+  }
+}
+
+/**
+ * Create a geometric outliers for each tile of a tileset when its geometry is loaded
```
valentinMachado, line 46: same review that the previous function
doc + use traverse

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
```diff
@@ -1,5 +1,133 @@
...
+ *
+ * @param {THREE.Object3D} tile  A 3DObject (should be a tile send by Tilesmanager.EVENT_TILE_LOADED)
+ * @param {number} threshOldAngle  An edge is only rendered if the angle (in degrees) between the face normals of the adjoining faces exceeds this value. default = 1 degree.
+ */
+export function appendWireframeToTileset(tile, threshOldAngle = 30) {
+  if (
+    tile.children[0] &&
+    tile.children[0].children[0] &&
+    tile.children[0].children[0].geometry &&
+    tile.children[0].children[0].geometry.isBufferGeometry
```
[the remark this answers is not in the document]
clementcolin: Because I think that point clouds (which can be loaded using 3DTiles) are not stored using a BufferGeometry

examples/3DTilesOutliers.html
clementcolin: done

examples/3DTilesOutliers.html
clementcolin: since 3DTiles are loaded, without elevation they will float making it awkward. It ended up by adding the elevation layer also

examples/3DTilesOutliers.html
clementcolin: Ty

clementcolin:
I just had a struck of genius (cc @valentinMachado ) : those methods should be in ThreeUtils as it can take generic three object as parameters

packages/browser/src/Component/Itowns/3DTiles/3DTilesUtils.js
clementcolin: done

examples/3DTilesWireframe.html
```diff
@@ -62,7 +70,7 @@
         for (const tilesManager of frame3DPlanar.layerManager.tilesManagers)
           tilesManager.addEventListener(
             udvizBrowser.TilesManager.EVENT_TILE_LOADED,
-            udvizBrowser.appendWireframeToTileset
+            udvizBrowser.THREEUtil.appendWireframeToTileset
```
valentinMachado, line 73: Since this is a three.js contextual function tileset should be replace by something like Object3D

valentinMachado (changes requested):
Every word outlier (doc + code) should be replace by wireframe in order to be consistent.
Since these methods belong to THREEUtil we are in a three.js context and tile or tilesmanager should not be quoted. Replace/remove everything related with tile.

In the second function _BATCHID should be passed as a parameter (for the same reason THREE does not know BATCHID concept) so its function becomes something like:
appendWireframeByGeometryAttributeToObject(obj,threshold,nameOfAttribute) <= this is quite long but explicit enough for an user of this function (addWireframeFromGeoAttrToObj <= short one)

packages/browser/src/Component/THREEUtil.js
```diff
@@ -269,39 +267,46 @@ module.exports = {
...
+   * This method is slower than appendWireframeToObject3D.
+   * @param {THREE.Object3D} object3D  An Object3D from three
    * @param {number} threshOldAngle  An edge is only rendered if the angle (in degrees) between the face normals of the adjoining faces exceeds this value. default = 1 degree.
+   * @param {string} nameOfGeometryAttribute The attribute used to split each geometry of the BufferGeometry
    */
-  appendWireframeByBatchIDToTileset: function (object3D, threshOldAngle = 30) {
+  appendWireframeByGeometryAttributeToObject3D: function (
+    object3D,
+    threshOldAngle = 30,
+    nameOfGeometryAttribute = '_BATCHID'
```
valentinMachado, line 298: I think this should not have a default parameter value since in three.js batch id means nothing

packages/browser/src/Component/THREEUtil.js
```diff
@@ -269,39 +267,46 @@ module.exports = {
...
+        !child.geometry.userData.hasWireframe
       ) {
         // This event can be triggered multiple times, even when the geometry is loaded.
-        // This bool avoid to create multiple outliers for one geometry
-        child.userData.hasOutlier = true;
+        // This bool avoid to create multiple wireframes for one geometry
+        child.userData.hasWireframe = true;
 
-        // Get the geometry that have the same _BATCHID to create its own outlier
+        // Get the geometry that have the same _BATCHID to create its own wireframe
```
valentinMachado, line 311: remove _BATCHID here

packages/browser/src/Component/THREEUtil.js
```diff
@@ -311,17 +316,21 @@ module.exports = {
         const batchid = new Array();
 
         // Iterate through each _BATCHID geometry
```
valentinMachado, line 320: same here remove _BATCHID

index.html
```diff
@@ -145,10 +145,19 @@
         'Billboard',
         'Frame3D',
         './examples/Billboard.html',
-        'This example implements a billboard (ie an embeded html content in a 3D view).',
+        'This example shows the outliers of the edges of a 3DTiles',
```
valentinMachado, line 148: remove outliers by wireframe

valentinMachado: this description is inverted with the one of billboard

packages/browser/src/Component/THREEUtil.js
```diff
@@ -240,4 +240,142 @@ module.exports = {
     camera.rotation.set(0, 0, -Math.PI / 2);
     camera.updateProjectionMatrix();
   },
+  /**
+   * Traverse a THREE.Object3D and append in each Object3D children a THREE.LineSegment geometry  representing its wireframe
+   * May be slow to create / load.
```
valentinMachado, line 245: what does this means, the tile load event is slow or this function is slow ?

packages/browser/src/Component/THREEUtil.js
```diff
@@ -240,4 +240,142 @@ module.exports = {
...
+            }
+            mesh.setIndex(indices);
+
+            // Create the wireframe geometry
+            const edges = new THREE.EdgesGeometry(mesh, threshOldAngle);
+
+            // Add this wireframe geometry to the global wireframe geometry
+            for (let l = 0; l < edges.attributes.position.count * 3; l++)
+              pos.push(edges.attributes.position.array[l]);
+            // Fill the _BATCHID buffer
```
valentinMachado, line 355: remove _BATCHID here

packages/browser/src/Component/THREEUtil.js
```diff
@@ -240,4 +240,142 @@ module.exports = {
...
+        // This bool avoid to create multiple wireframes for one geometry
+        child.userData.hasWireframe = true;
+
+        // Get the geometry that have the same _BATCHID to create its own wireframe
+        let startIndex = 0;
+
+        // Position array that will be filled with each geometry ordered by _BATCHID
+        const pos = new Array();
+
+        // BatchID array that will be filled with _BATCHID
```
valentinMachado, line 317: remove _BATCHID

valentinMachado (commented):
need some renaming

valentinMachado:
oups I resolve conflict through the github interface and it has merged master into this branch...

valentinMachado (approved):

PR #579, opened 2023-03-22
568 repair html layerchoice + document

valentinMachado:


packages/browser/src/Component/HTMLUtil.js
```diff
@@ -49,3 +51,90 @@ export function clearChildren(element) {
...
+ * @returns {{parent:HTMLElement,container:HTMLElement}} - parent is the element to add container is the element to fill
+ */
+export function createDisplayable(label) {
+  const parent = document.createElement('div');
+  const displayableButton = document.createElement('button');
+  parent.appendChild(displayableButton);
+  const container = document.createElement('div');
+  parent.appendChild(container);
+
+  let hidden = true; // true by default
```
valentinMachado, line 68: this bool is useless

packages/browser/src/Component/HTMLUtil.js
```diff
@@ -49,3 +51,90 @@ export function clearChildren(element) {
...
+ */
+export function createDateIntervalInput(labelText) {
+  const parent = document.createElement('div');
+
+  const label = document.createElement('label');
+  label.innerText = labelText;
+  parent.appendChild(label);
+
+  // start date interval
+  const startFromDiv = document.createElement('div');
```
valentinMachado, line 120: could be factorize with a function createLabelInputDate(labelText)

packages/browser/src/Component/Widget/Server/Document/Core/View/DocumentNavigatorWindow.js
```diff
@@ -71,57 +95,90 @@ export class DocumentNavigatorWindow {
...
+      this.documentListContainer = document.createElement('div');
+      this.rootHtml.appendChild(this.documentListContainer);
+      {
+        // label document count
+        const labelDocCount = document.createElement('h3');
+        this.documentListContainer.appendChild(labelDocCount);
+        {
+          this.docCountElement = document.createElement('span');
+          labelDocCount.appendChild(this.docCountElement);
+          labelDocCount.innerText = 'Document(s)';
```
valentinMachado, line 113: this line should be a scope above

packages/browser/src/Component/Widget/Server/Document/Link/View/DocumentLinkInterface.js
```diff
@@ -218,43 +263,11 @@ export class DocumentLinkInterface {
   // ////////////
   // //// GETTERS
 
-  get linkListId() {
```
valentinMachado, line 221: cannot be deleted because used bu getters left

packages/browser/src/Component/Widget/Server/Document/Link/View/DocumentLinkInterface.js
```diff
@@ -4,7 +4,11 @@ import { DocumentProvider } from '../../Core/ViewModel/DocumentProvider';
 import { Link } from '../Model/Link';
 import { LinkProvider } from '../ViewModel/LinkProvider';
 
-import { findChildByID } from '../../../../../HTMLUtil';
+import {
+  createDisplayable,
+  createLabelInput,
+  findChildByID,
+} from '../../../../../HTMLUtil';
```
clementcolin, line 11: For my knowledge (and improve my JS competence) : since HTMLUtils is exported in index.js, is it possible to make the import from it ? 
This means that if we move HTMLUtil one day, we will only have to change the path in index.js (+ this can improve readability)

clementcolin (approved):

packages/browser/src/Component/Widget/Server/Document/Link/View/DocumentLinkInterface.js
valentinMachado: Yes I was thinking about that, one solution is what you say but we are loosing some information about dependencies between "modules"/subfolder since every file (theoratically) depends of index.js (aka @ud-viz/browser).
I was thinking about another way a bit more complex but allowing a good readability about dependencies which is each file can import module from the first api file up (for this file it would be [this one](https://github.com/VCityTeam/UD-Viz/blob/master/packages/browser/src/Component/Widget/Server/Document/Document.js)) and so on.

Not sure to be clear but we could create an issue about it

Issue #537, opened 2023-02-24
Explain the target code organisation wished


valentinMachado:
Each UD-Viz package [directory hierarchy](https://github.com/VCityTeam/UD-Viz/tree/master/packages/browser#directory-hierarchy) together with its classes hierarchy/organization derives from "some" organizational logic realizing the synthesis between different code design forces/principles. Among such forces lies e.g.
* the distinction between front-end/back-end/what-is-shared-by-both aka [browser/node/shared](https://github.com/VCityTeam/UD-Viz/tree/master/packages)
* the distinction between notions like [component/extension/module/widget/view/model/controler](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/Definitions.md)
* design patterns like [Model View Controler](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/ArchitectureMVCTargetDesign.md) with notions like `provider`, `controler`, `service`
* probably some undocumented/implicit/too-well-known-to-be-explicit/unconscious design principle/patterns ?

Make such organizational forces explicit and clarify the resulting organizational logic behind UD-Viz (like UD-SV proposes at the [repository level](https://github.com/VCityTeam/UD-SV/blob/master/ImplementationKnowHow/ProjectOrganizationalLogic/Readme.md)).
In doing so, make sure to distinguish the internal organizational logic from the package end-user usage logic (for whom the package API hides away the implementations details  and only reveals the usage logic) .

## Related issues
* Should we expose the 'providers' (in UD-Viz API) or should they remain in the 'modules' ? (refer to [this weekly discussion](https://github.com/VCityTeam/VCity/blob/master/Weekly_Minutes/2023/Weekly_Minutes_2023_02_24.md#discussions)) 
* [Issue #538](https://github.com/VCityTeam/UD-Viz/issues/538) 
* [Issue #512: Separate Window GUI component from widget](https://github.com/VCityTeam/UD-Viz/issues/512)

valentinMachado:
```mermaid
flowchart LR
AssetManager
ExternalGame
Frame3D
Itowns
Widget
FileUtil
HTMLUtil
InputManager
RequestAnimationFrame
SocketIOWrapper
THREEUtil
AssetManager --> THREEUtil
Frame3D --> HTMLUtil
Frame3D --> Itowns
Frame3D --> THREEUtil
Itowns --> Widget
Widget --> Itowns
Widget --> InputManager
ExternalGame --> AssetManager
[...]

PR #585, opened 2023-03-24
583 move temporal model to itowns folder

valentinMachado:
motivation discussed here => #537 
will depend of this discussion => #587


valentinMachado:
The 3DTilesTemporal model should be stay an extension for now => we should modify the addlayerfrom config to just generic 3DTilesTemporal and this is to the user to register extension before loading 3DTiles

PR #606, opened 2023-04-07
Update Readme.md

EricBoix:
A simple badge for the landing page.


valentinMachado:
This badge refers to an old version of ud-viz (when it was a single package) now ud-viz is an [organization](https://docs.npmjs.com/organizations) with multiple packages [see](https://www.npmjs.com/~vcity).

PR #607, opened 2023-04-11
fix(inputmanager): capslock insensitive

mathieuLivebardon:
https://github.com/VCityTeam/UD-Imuv/issues/399


clementcolin (approved):

packages/browser/src/Component/InputManager.js
```diff
@@ -16,8 +16,10 @@ const Command = require('@ud-viz/shared').Command;
 export class InputManager {
   /**
    * Manage user inputs with a poll system (https://en.wikipedia.org/wiki/Polling_(computer_science))
+   *
+   * @param {boolean} [isCaseSensitive=false] true so that the inputs are case sensitive
    */
-  constructor() {
+  constructor(isCaseSensitive = true) {
```
valentinMachado, line 22: jsdoc default param is false and the default value in constructor is true. According what you said looks like the default behavior should be false.

valentinMachado (changes requested):

valentinMachado:
I push an example of a bug. Go to http://localhost:8000/examples/bug.html.
When you type a you trigger the two events as expected but if you type A none are trigger

packages/browser/src/Component/InputManager.js
mathieuLivebardon: Good point

valentinMachado (approved):

PR #608, opened 2023-04-11
fix(slideshow): remove correctly listeners fix #603

### LINKED ISSUE #603: bug(slideshow)
In sidebarwidget open and close slideshow widget. This error is throw

```
SlideShow.js:146 Uncaught TypeError: Failed to execute 'removeEventListener' on 'EventTarget': 2 arguments required, but only 1 present.
    at SlideShow.removeListeners (SlideShow.js:146:1)
    at SlideShow.dispose (SlideShow.js:165:1)
    at sideBarButton.onclick (SideBarWidget.js:707:1)
```

mathieuLivebardon:


valentinMachado:
In SideBarWidget if you open Slideshow then close then open again console throw this error:

```
Uncaught TypeError: Cannot read properties of null (reading 'parentElement')
    at sideBarButton.onclick (SideBarWidget.js:711:1)
```

packages/browser/src/Component/Widget/SlideShow/SlideShow.js
```diff
@@ -593,7 +593,7 @@ export class SlideShow {
    */
   initInputListener() {
     this.hidePlaneListener = (event) => {
-      if (event.key != 'h') return;
+      if (event.key.toLowerCase() != 'h') return;
```
clementcolin, line 596: This may be a dumb question, but since we have an input manager (and a PR about case sensitivity), should we not handle case sensitivity here also?

valentinMachado: Originally slideshow was working with the InputManager, but since our widgets are meant to be contributed to the itowns project we have to make them not dependant of other part of our code (I doubt that the InputManager will be part of the itowns project).

valentinMachado (approved):

PR #612, opened 2023-04-12
584 clean itowns folder

### LINKED ISSUE #407: Point Cloud parameter pc_size not working

### LINKED ISSUE #583: Move Temporal Model to Itowns folder
see motivation here => #537

### LINKED ISSUE #584: Clean Itowns folder
This issue aim to clean the folder [packages/browser/src/Component/Itowns](https://github.com/VCityTeam/UD-Viz/tree/master/packages/browser/src/Component/Itowns). The code in that folder is used by our [Widget](https://github.com/VCityTeam/UD-Viz/tree/master/packages/browser/src/Component/Widget) and propose some features to itowns framework, in order to contribute our widget in [itowns/widget](https://github.com/iTowns/itowns/tree/master/src/Utils/gui) we have to:
* add these features to itowns, if they are relevant/necessary
* refacto our widget using exclusively itowns, if they are not 

In anycase this folder should be "empty", or at least be considered as a "buffer of code" the time to mature it and to add it to itowns. Note that we could have a folder ItownsUtil where we could have code as [AddLayerFromConfig](https://github.com/VCityTeam/UD-Viz/blob/master/packages/browser/src/Component/Itowns/AddLayerFromConfig.js) which should not be in itowns (and not in our Widget) but is very useful in an @ud-viz/browser context (to create application). 

As there are many code in that folder, I propose to open an issue for each folder/file in that folder, to discuss them one by one.

* [ ] #586
* [ ] #587
* [x] #588
* [x] #589
* [x] #590

This thread is to keep a global track of this work. And to define where are priorities.

related #537

### LINKED ISSUE #586: Move AddLayerFromConfig to ItownsUtil
As AddLayerFromConfig is a module to quickly configure itowns layers from a config file, we could consider it as an Util in an @ud-viz/browser context.

### LINKED ISSUE #587: Clean TemporalModel
Is related to [this PR](https://github.com/VCityTeam/UD-Viz/pull/585).

Apparently the 3DTilesTemporal model can be registered as an extension in itowns [ref](https://github.com/VCityTeam/UD-Viz/blob/dba3485315a3a5a11e184f67505f93fb6144f556/packages/browser/src/Component/Itowns/AddLayerFromConfig.js#L30)

What are the motivation to do so and not implements this model in itowns ?

If this model is not meant to be implements in itowns, we should then rework this [function](https://github.com/VCityTeam/UD-Viz/blob/dba3485315a3a5a11e184f67505f93fb6144f556/packages/browser/src/Component/Itowns/AddLayerFromConfig.js#L17) to not be dependant of the 3DTilesTemporal model

### LINKED ISSUE #588: Clean LayerManager
[LayerManager documentation](https://github.com/VCityTeam/UD-Viz/blob/583-move-temporal-model-to-itowns-folder/docs/static/Devel/UD_Viz_Browser/LayerManager/layerManager.md)

### LINKED ISSUE #589: Move Itowns/Component to ItownsUtil

### LINKED ISSUE #590: Clean 3DTiles
This folder implements the notion of 3DTiles in @ud-viz/browser, apparently itowns has its own definition see #342. As there are many files in that folder, I split into differents issues:

* [ ] #591
* [ ] #592
* [x] #593
* [x] #594
* [x] #595
* [ ] #596

### LINKED ISSUE #591: Clean TilesManager
[TilesManager documentation](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/UD_Viz_Browser/3DTiles/TilesManager.md)

related #588

### LINKED ISSUE #592: Clean StyleManager
[StyleManager documentation](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/UD_Viz_Browser/3DTiles/StyleManager.md)

related #591 #588 

References:
  - [Symbology encoding de l'OGC](https://docs.ogc.org/is/18-067r3/18-067r3.html)
  - [3D Styles styling](https://cesium.com/learn/cesiumjs-learn/cesiumjs-3d-tiles-styling/)
  - [mapbox styles](https://docs.mapbox.com/api/maps/styles/)
  - [iTowns' Style](https://github.com/iTowns/itowns/blob/master/src/Core/Style.js#L132)

### LINKED ISSUE #593: Clean/Move 3DTilesUtils
[3DTilesUtil documentation](https://github.com/VCityTeam/UD-Viz/blob/master/docs/static/Devel/UD_Viz_Browser/3DTiles/3DTilesUtils.md)

### LINKED ISSUE #594: Clean CityObject
This class should be the first thing to discuss, as many of the other issue of #590 and thus #584 are related to it.
Basically it's record an interval in the vertex array of the 3DTile (or a batch id) to attach some semantic to it (props which are stored in the batchtable)

For me it's not relevant but I might miss some use case so do not hesiate to expose your opinion over there.

## References
- Corresponding [iTowns discussion](https://github.com/iTowns/itowns/discussions/2054)

### LINKED ISSUE #595: Clean CityObjectStyle
Why not using [THREE.Material](https://threejs.org/docs/index.html?q=material#api/en/materials/Material) ?

related #592

valentinMachado:
Delete Itowns folder (move what's needed to be keep in [ItownsUtil](https://github.com/VCityTeam/UD-Viz/blob/584-clean-itowns-folder/packages/browser/src/Component/ItownsUtil.js))
Delete Link of document
> Reason was it was depending on an attribute in the batchTable "cityobject.database.id" which is not present in our 3DTiles anymore, if in the futur if we want this feature again we should reimplement it the right way meaning: create a 3DTilesExtension to guaranteed that this "cityobject.database.id" is present in the tileset/batchTable.

Delete CityObjectModule
> itowns equivalent [C3DTFeature](https://github.com/iTowns/itowns/blob/master/src/Core/3DTiles/C3DTFeature.js)

[Delete Debug3DTiles](https://github.com/iTowns/itowns/discussions/2061)

Depends of this [PR](https://github.com/iTowns/itowns/pull/2068) and this [one](https://github.com/iTowns/itowns/pull/2060)

Temporal has been quite a lot of refacto documentation is not up to date [issue related](https://github.com/VCityTeam/UD-Viz/issues/616)

close #594 
close #595 
close #593 
close #592 
close #591 
close #590 
close #589 
close #588 
close #587 
close #586 
close #584 
close #583 
close #407


LorenzoMarnat:
Since you've deleted the CityObject widget, is there a way to do the same (click on a feature, apply a material and display its attributes) with the new implementation in iTowns ? Do you you think it will be hard to create an example of this (maybe in a future PR) ?

valentinMachado:
> Since you've deleted the CityObject widget, is there a way to do the same (click on a feature, apply a material and display its attributes) with the new implementation in iTowns ? Do you you think it will be hard to create an example of this (maybe in a future PR) ?

I implement an example usage in SideBarWidget [here](https://github.com/VCityTeam/UD-Viz/blob/2ba3e18d549a72d257ae63ef42cf57b508b80929/packages/browser/src/SideBarWidget/SideBarWidget.js#L37) and [there](https://github.com/VCityTeam/UD-Viz/blob/2ba3e18d549a72d257ae63ef42cf57b508b80929/packages/browser/src/SideBarWidget/SideBarWidget.js#L166) but you right it should be a dedicated example ([issue](https://github.com/VCityTeam/UD-Viz/issues/621))

index.html
```diff
@@ -238,10 +221,10 @@
...
 
       addProjectElement(
-        'GeoJSON',
+        'Add Layers from config',
         'Frame3D',
-        './examples/GeoJson.html',
-        `This example visualizes several GeoJSON datasets (such as bus lines, Velo'v stations, and humidity zones) alongside 3D planar data.`,
+        './examples/AddLayersFromConfig.html',
+        `This example show how to add layers (such as geojson, elevation, base map, label, 3DTiles).`,
         './examples/assets/img/mosaic/GeoJSON.png'
```
clementcolin, line 228: Maybe rename this png to match the example

packages/browser/src/Component/ItownsUtil.js
```diff
@@ -1,68 +1,29 @@
...
 
 /**
  * It creates a 3D Tiles layer,
  * and adds it to the layer manager
  *
  * @param {*} layer - the layer object from the config file
- * @param  {LayerManager} layerManager - the layer manager object
  * @param {itowns.View} itownsView - the itowns view
+ * @param {object} extensions - optional extensions
  * @returns {itowns.C3DTilesLayer} A 3D Tiles Layer
```
clementcolin, line 16: the return seems to be deprecated

packages/browser/src/Component/ItownsUtil.js
```diff
@@ -71,23 +32,9 @@ export function setup3DTilesLayer(layer, layerManager, itownsView) {
...
     itownsView
   );
-  if (overrideMaterial) {
-    $3dTilesLayer.overrideMaterials = material;
-    $3dTilesLayer.material = material;
-  }
-
-  const $3DTilesManager = new TilesManager(itownsView, $3dTilesLayer);
-
-  if (layer['color']) {
```
clementcolin, line 85: Deleting those lines make it impossible to change the color of a tileset using the config file : is it a feature that we want to drop ? 
Same remark about point size of point cloud

packages/browser/src/Component/ItownsUtil.js
```diff
@@ -96,26 +43,20 @@ export function setup3DTilesLayer(layer, layerManager, itownsView) {
...
- * @param {LayerManager} layerManager a layer manager
  * @param {itowns.View} itownsView - the itowns view
- * @returns {Map<itowns.C3DTilesLayer>} a map of each 3d tiles layer
  */
-export function add3DTilesLayers(
-  config3DTilesLayers,
-  layerManager,
-  itownsView
-) {
+export function add3DTilesLayers(config3DTilesLayers, itownsView) {
```
clementcolin, line 48: As this method is not really that relevant,  don't you think that we can drop this method and do the loop on the layers directly in the code ?

packages/browser/src/Component/ItownsUtil.js
```diff
@@ -396,3 +337,72 @@ export function addElevationLayer(configElevationLayer, itownsView, extent) {
...
+  if (!layer.isC3DTilesLayer) return;
+
+  const coordinates = itownsView.camera.position();
+  const extent = layer.extent;
+  coordinates.x = (extent.east + extent.west) / 2;
+  coordinates.y = (extent.north + extent.south) / 2;
+  coordinates.z = 200;
+  if (layer.tileset.tiles[0])
+    coordinates.z = layer.tileset.tiles[0].boundingVolume.box.max.z;
+  focusCameraOn(itownsView, itownsView.controls, coordinates, {
```
clementcolin, line 404: maybe vertical and horizontal distance can be also set as parameter ?
In the futur, we may also want to compute them to exactly match the extent of the 3DTiles

packages/browser/src/Component/Widget/LayerChoice/views/LayerChoice.js
```diff
@@ -66,7 +64,9 @@ export class LayerChoice {
   innerContentColorLayers() {
     const list = this.colorLayersSpoilerBoxElement;
     list.innerHTML = '';
-    const layers = this.layerManager.getColorLayers();
+    const layers = this.itownView
```
clementcolin, line 67: Method such as get color layers and so on can be utils contributed to itowns or method in this class ?

packages/browser/src/Component/Widget/LayerChoice/views/LayerChoice.js
```diff
@@ -107,18 +104,17 @@ export class LayerChoice {
   innerContentElevationLayers() {
     const list = this.elevationLayersSpoilerBoxElement;
     list.innerHTML = '';
-    const layers = this.layerManager.getElevationLayers();
+    const layers = this.itownView
```
clementcolin, line 107: same comment

packages/browser/src/Component/Widget/LayerChoice/views/LayerChoice.js
```diff
@@ -135,17 +131,31 @@ export class LayerChoice {
   innerContentGeometryLayers() {
     const list = this.geometryLayersSpoilerBoxElement;
     list.innerHTML = '';
-    const layers = this.layerManager.getGeometryLayers();
+    const layers = this.itownView
```
clementcolin, line 134: dito

packages/browser/src/Component/Widget/Server/SPARQL/View/SparqlQueryWindow.js
```diff
@@ -146,45 +129,81 @@ export class SparqlQueryWindow extends EventSender {
...
-      );
-      if (cityObject) {
-        focusCameraOn(
-          this.layerManager.view,
-          this.layerManager.view.controls,
-          cityObject.centroid,
-          {
-            verticalDistance: 200,
-            horizontalDistance: 200,
+    const fetchC3DTileFeatureWithNodeText = (node_text) => {
```
clementcolin, line 132: could be a method in itownsutils to find a feature in a 3DTiles using a feature name (here 'gml_id') and a value (here node_text)

clementcolin: note that it is possible to find multiple matching feature

packages/browser/src/Component/ItownsUtil.js
valentinMachado: I agree even more we may want to customize the for loop creation by adding a style for example. In my opinion all addLayerFromConfig should be removed because the itowns API is enough.

If everyone agree I will remove these functions ?

packages/browser/src/Component/ItownsUtil.js
valentinMachado: To be honest I leave them the way they were because we should contribute this feature in itowns => [discussion](https://github.com/iTowns/itowns/discussions/2075)

packages/browser/src/Component/Widget/LayerChoice/views/LayerChoice.js
valentinMachado: The itowns API to select certain layer is [view.getLayers(filterFunction)](https://github.com/iTowns/itowns/blob/11582c967ca22ba9fa15ffefba75506e04759990/src/Core/View.js#L440)
which is the itowns way I agree with that since I rather have one function (getLayers) with x parameters (filter layer) than x * function with x the number of typed layer which keep a low number of entry point. As you can noticed I am not using the parameter but .filter method of an array in javascript which is question of taste.

packages/browser/src/Component/Widget/LayerChoice/views/LayerChoice.js
valentinMachado: see previous answer

packages/browser/src/Component/Widget/LayerChoice/views/LayerChoice.js
valentinMachado: same

packages/browser/src/Component/Widget/Server/SPARQL/View/SparqlQueryWindow.js
valentinMachado: Actually it's looks like a [.filter on an array](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) here features are stored within the layer in Map<tileId, Map< batchId, C3DTFeature>> so if tilesC3DTFeatures is that object writing something like ([according stackoverflow](https://stackoverflow.com/questions/48707227/how-to-filter-a-javascript-map))

```
for ( const [tileId, tileC3DTFeatures] of tilesC3DTFeatures){
 const mapC3DTFeatureFiltered = new Map(
  [...tileC3DTFeatures]
  .filter(([batchId, c3DTFeature]) => //return tru when you want feature in the result )
 );
}
```

should do the work don't think we need an util for that, I didn't write it that way though (I just lean how to do it ^^)

packages/browser/src/Component/ItownsUtil.js
valentinMachado: For the point cloud I will test the fresh new example and see if nothing is broken with the new implem. One thing I am thinking about is that overrideMaterials is now deprecated in itowns in order to force user to use style but I am not sure that style is compatible with point cloud and I need to see if it could if not I 'll just ask to itowns to restore the overrideMaterials attribute for the long term for that case (ie point cloud)

packages/browser/src/Component/ItownsUtil.js
valentinMachado: It's actually up to date, createC3DTilesLayer, previous setupC3DTilesLayer now return a layer

packages/browser/src/Component/ItownsUtil.js
clementcolin: Style in json file will come back with itowns then, fine for me.

packages/browser/src/Component/ItownsUtil.js
clementcolin: I'm fine with this.

I guess code examples  are always nice to have around, so go for it if you want

packages/browser/src/Component/ItownsUtil.js
clementcolin: Ok

clementcolin (approved):

valentinMachado:
I ve created an [issue](https://github.com/VCityTeam/UD-Viz/issues/631) about the code that should be moved from src to examples/assets/js

PR #630, opened 2023-04-24
Add the reload package to automatically refresh tabs when there are modifications

mathieuLivebardon:
WARNING this PR is based on #627 

This PR link a reload function to dev workflow.
- Adds a new package : [reload](https://github.com/alallier/reload)
- Adds in ./bin/Constant : RELOAD_PORT is the port where reload server run
- Add in all examples a string (SCRIPT_TAG_RELOAD) which is replaced with express string replace by a script tag that the reload package needs.


packages/node/package.json
```diff
@@ -24,11 +24,11 @@
...
     "child-process-promise": "^2.2.1",
     "express": "^4.18.2",
+    "gm": "^1.23.1",
+    "pngjs": "^6.0.0",
     "puppeteer": "^19.4.1",
     "socket.io": "^4.5.4",
-    "string-replace-middleware": "^1.0.2",
-    "gm": "^1.23.1",
-    "pngjs": "^6.0.0"
+    "string-replace-middleware": "^1.0.2"
```
valentinMachado, line 31: string replace should not be a depency of node package since it's only use in the mono repo context

packages/node/src/ExpressAppWrapper.js
valentinMachado, line 1: this file has been removed on master branch [see](https://github.com/VCityTeam/UD-Viz/pull/627)

valentinMachado, line 1: this file has been removed on master branch [see](https://github.com/VCityTeam/UD-Viz/pull/627)

valentinMachado (approved):

PR #637, opened 2023-05-02
refact(examples + Thread) some renaming + no bundle builded

valentinMachado:
No bundle is builded to launch examples game thread child (previous default-thread).
I also rename all "default" by "example" to make thing clearer and use a minimal amount of word.
I rename "host" by "examplesBackEnd" to introduce the word example, maybe "examplesHost" is more appropriate I don't know.

This was also the opportunity to refactor the `Thread` class in node package. I delete `ThreadProcessRoutine` and move the code in `Thread`. I introduce notion of `Parent` and `Child` which imo is more clearer as well.

I run an [npm update](https://docs.npmjs.com/cli/v6/commands/npm-update) which has updated package.lock, I could undo it but imo that's a good thing (especially when no error is encountered), a warning has been triggered by puppetter though that's why `headless: true` => `headless: new`


docs/static/Devel/Developers.md
```diff
@@ -61,7 +61,7 @@ consists in
...
-| `npm run host`           | host the bundle with an [ExpressAppWrapper](../../../bin/host.js) + a default game socket service. <br>http://locahost:8000/                                                                                                                                                                                    |
+| `npm run examples-back-end`           | run an http server (with some [string-replace](https://www.npmjs.com/package/string-replace-middleware))  + a game socket service. <br>http://locahost:8000/                                                                                                                                                                                    |
 | `npm run start`          | Run `npm run build-browser` and `npm run host`                                                                                                                                                                                                                        |
```
mathieuLivebardon, line 65: This no longer true.

valentinMachado:
> I am confusing by the naming and the order of the words of the scripts `examples-back-end` and files `examplesBackEnd.js`, `examplesGameThreadChilds.js` . For my opinion we have to keep a consistance with `debugExamples.js`.

You mean we should call them `gameThreadChildExamples.js` and `backEndExamples.js` ?

docs/static/Devel/Developers.md
valentinMachado: What is no longer true ? actually this is what is doing the back end of the examples

mathieuLivebardon (approved):
🏗️

Issue #312, opened 2022-03-11
Create a 2D visualization mode


clementcolin:
Not all 3D data have an altitude, meaning that some are "flying" and some are not when not using the elevation layer.

It would be nice to provide a feature to "put on the ground" each tileset, or set their altitude at 0,

valentinMachado:
How do you detect that a 3D data (<= 3DTiles you mean ?) does not have an altitude ?

valentinMachado:
Ok so if I understand well, an object3D (3DTiles or THREE.Object3D) does not have an altitude if the lower point of the object3D is above the higher point of the elevation layer (in the xy square of the object3D) right ?

And if it does not have an altitude then we translate the object3D on the z axis of obj3d_lower_point.z - elevation_layer_higher_point.z something like that ?

I think this will works for object3D on a flat ground.

![dessin](https://user-images.githubusercontent.com/14295096/233137980-d0749311-f94c-428a-ba61-f40f4f765bba.png)

here you can first see my art skill and take a moment to appreciate

then in that case the building in red (lower point in yellow) and the ground in gray (higher point in green) are not matching the "flying" condition and as you can see we could translate this building to "put it on the ground"

one approach could be to divide the area in several sub area and to operate the same algorithm (as we could consider the more we divide the more the ground will be flat) then do an average of the translation amount on the z axis.

Another thing I am thinking about is as 3DTiles wrapped their own transformation we don't want to move them..
Typically we move the 3DTiles everything is fine, the 3DTiles unload the 3DTiles reload and now it's flying again.

So I guess the right approach would be to move the elevation layer (I don't know how without deleting recreating the layer but should be manageable)

valentinMachado:
That's explain why I was still not getting the title ^^

[...]

PR #638, opened 2023-05-03
Create a 2d visualization mode

valentinMachado:
I've made an example illustrating how to flatten `C3DTFeature` of the 3DTiles. 
One thing we have not think about [there](https://github.com/VCityTeam/UD-Viz/issues/312) is that the boundingVolume has to be updated as well since we want the 3DTiles to be cull/subdivide according the right boundingVolume (box/sphere).

This PR is just to "POC" the concept then we could think about contributing this feature to itowns, maybe by passing an optional 
boolean at the construction of the `C3DTilesLayer`. And if not we could think how to integrate it in @ud-viz/browser.


packages/browser/src/ItownsUtil.js
```diff
@@ -415,3 +415,37 @@ export function focusC3DTilesLayer(itownsView, layer) {
     horizontalDistance: 200,
   });
 }
+
```
valentinMachado, line 418: I think we don't need these functions. If you want to apply some treatment to tiles with geometry I rather doing something like that:

```js
layer.object3D.traverse((child) => {
 if(!isNaN(child.tileId) && child.tileId >= 0 && child.tileId < layer.tileset.tiles.length ){
  // here child is a tile with geom and tileId
 }
});



```

valentinMachado:
Looks like there is some problem with the bouding volume 
2D Visualization example

clementcolin:
> Looks like there is some problem with the bounding volume 2D Visualization example

I think that i will show the bounding volume in this example to actually visualize what is hapenning when changing visualization mode

examples/2DVisualization.html
```diff
@@ -0,0 +1,176 @@
...
+                  child.tileId >= 0 &&
+                  child.tileId < layer.tileset.tiles.length &&
+                  layer.tilesC3DTileFeatures.has(child.tileId)
+                ) {
+                  for (const [
+                    // eslint-disable-next-line no-unused-vars
+                    batchId,
+                    c3DTFeature,
+                  ] of layer.tilesC3DTileFeatures.get(child.tileId)) {
+                    child.traverse((child) => {
```
valentinMachado, line 93: child is defined twice

examples/2DVisualization.html
```diff
@@ -0,0 +1,176 @@
...
+                            child.geometry.attributes.position.array[
+                              index + 2
+                            ] -= worldBox3.min.z - featureElevation; // translation
+                          }
+                          child.geometry.attributes.position.needsUpdate = true;
+                        });
+                      }
+                    });
+                  }
+                  child.traverse((el) => {
```
valentinMachado, line 137: you could optimize this compution by computing min and max z when offseting features then only modify child.boundingVolume.box.max.z and .min.z

examples/2DVisualization.html
```diff
@@ -0,0 +1,176 @@
...
+        frame3DPlanar.appendToUI(button);
+
+        const planarLayer = frame3DPlanar.itownsView.tileLayer;
+        const elevationLayer = frame3DPlanar.itownsView
+          .getLayers()
+          .filter((el) => el.isElevationLayer)[0];
+
+        elevationLayer.scale = 0;
+
+        const change3DTilesHeight = () => {
```
valentinMachado, line 76: the name of this function should be more an update that a change since it can be applied to a 3DTiles without changing anything

examples/2DVisualization.html
```diff
@@ -0,0 +1,176 @@
...
+
+        button.onclick = changeVisualizationMode;
+
+        frame3DPlanar.itownsView
+          .getLayers()
+          .filter((el) => el.isC3DTilesLayer)
+          .forEach((layer) => {
+            layer.addEventListener(
+              udvizBrowser.itowns.C3DTILES_LAYER_EVENTS.ON_TILE_CONTENT_LOADED,
+              change3DTilesHeight
```
valentinMachado, line 169: here you should update 3DTiles elevation only for the one which has loaded

examples/2DVisualization.html
```diff
@@ -0,0 +1,176 @@
...
+
+        const button = document.createElement('button');
+        button.innerHTML = 'Change visualization mode to 3D';
+        button.style.zIndex = 2;
+        button.style.position = 'relative';
+        button.style.cursor = 'pointer';
+        frame3DPlanar.appendToUI(button);
+
+        const planarLayer = frame3DPlanar.itownsView.tileLayer;
+        const elevationLayer = frame3DPlanar.itownsView
```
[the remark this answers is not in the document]
clementcolin: done in 
[fd63ea7](https://github.com/VCityTeam/UD-Viz/pull/638/commits/fd63ea7f327166f6528181bb1d9f8d8e44ae1445)

examples/2DVisualization.html
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                  });
+                }
+              });
+            });
+          frame3DPlanar.itownsView.notifyChange();
+        };
+
+        const changeVisualizationMode = () => {
+          const elevationLayer = frame3DPlanar.itownsView
```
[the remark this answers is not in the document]
clementcolin: done in [fd63ea7](https://github.com/VCityTeam/UD-Viz/pull/638/commits/fd63ea7f327166f6528181bb1d9f8d8e44ae1445)

examples/2DVisualization.html
```diff
@@ -10,166 +10,170 @@
...
+              tileContent.tileId >= 0 &&
+              tileContent.layer.tilesC3DTileFeatures.has(tileContent.tileId)
+            ) {
+              tileContent.traverse((el) => {
+                for (const [
+                  // eslint-disable-next-line no-unused-vars
+                  batchId,
+                  c3DTFeature,
+                ] of tileContent.layer.tilesC3DTileFeatures.get(tileContent.tileId)) {
+                  if (el.geometry && el.geometry.attributes._BATCHID) {
```
valentinMachado, line 90: this if condition should be tested before the for loop

examples/2DVisualization.html
```diff
@@ -10,166 +10,170 @@
...
-            .filter((el) => el.isElevationLayer)[0];
-
-          elevationLayer.scale = elevationLayer.scale == 1 ? 0 : 1;
-          button.innerHTML =
-            'Change visualization mode to ' +
-            (elevationLayer.scale == 1 ? '2D' : '3D');
-          change3DTilesHeight();
-        };
+            }
+            // });
```
valentinMachado, line 142: these comments should be removed

valentinMachado:
@clementcolin I moved some variable declaration + assignement + add some comment. If you are fine with my commit we could merge this

clementcolin (approved):

PR #639, opened 2023-05-03
Create a widget to display informations of the c3dtfeature + add/remove dynamically 3DTiles

### LINKED ISSUE #187: Add 3DTiles layer from URL in a demo
Adding 3DTiles layer in a running demo from an URL could be a useful feature. For example, we could propose in a menu a selection of 3DTiles tilesets, like one for each city, and the user can chose the ones he wants to see

valentinMachado:
Display info of a `C3DTFeature`
Toggle visibility + Add + Remove dynamically `C3DTilesLayer`

close #187


packages/browser/src/Component/Widget/C3DTiles.js
```diff
@@ -0,0 +1,174 @@
+import * as itownsWidget from 'itowns/widgets';
+import * as itowns from 'itowns';
+import { createLabelInput } from '../HTMLUtil';
+import * as THREE from 'three';
+
+const DEFAULT_OPTIONS = {
+  position: 'top-right', // should be deprecated https://github.com/iTowns/itowns/issues/2005
+};
+
+export class C3DTiles extends itownsWidget.Widget {
```
clementcolin, line 10: This make me wonder : As we are now really close to Itowns, should we not always create widget extending itownsWidget ?

valentinMachado: yes we should to make a possible contribution easier then, even if this class is quite doing nothing.. I tried with this widget to illustrate how should looks like a widget implementation.

LorenzoMarnat:
I can only select a building once (then clicking on a building doesn't select it)

image

Also, I can't drag or zoom in the view, is this a bug or a feature ?

It's really cool to load tilesets directly in the widget via URL tho

LorenzoMarnat:
It seems that if I remove some tilesets, I can use UD-Viz correctly. Maybe my browser can't handle 9 tilesets

image

valentinMachado:
> I can only select a building once (then clicking on a building doesn't select it)
> 
> image

What should be the behavior ? 
It should be possible to select severals features and to unselect the feature has to be clicked again ?

valentinMachado:
> It seems that if I remove some tilesets, I can use UD-Viz correctly. Maybe my browser can't handle 9 tilesets
> 
> image

The widget css is not applied, it should looks like the image below. My guess is that you can't use controls because the widgets domElement is covering the screen catching event. I don't know why though, can you check css import on your side ?(I check the branch and the css was well loaded)

Capture d’écran du 2023-05-05 13-53-19

LorenzoMarnat:
Don't know what changed but it works. Maybe my browser was using some cached CSS

image

LorenzoMarnat (approved):

PR #650, opened 2023-05-10
refact(Thread) runChildProcess is a method of Child

valentinMachado:
That was a bit weird to have runChildProcess out of Child class so I made a method and quite modify the api.
I rather dispatch an event ON_GAME_CONTEXT_LOADED instead of returning a Promise with the run method.

I also rename EVENT and KEY which are now MESSAGE_EVENT and MESSAGE_KEY where message is used to define what is send between child and parent in Thread.

Nothing functionally should have changed that's just to clarify stuff.

### DIFF
--- a/packages/node/src/Game/Thread.js
+++ b/packages/node/src/Game/Thread.js
@@ -199,7 +199,7 @@ class Child {
    *
    * @param {Object<string,Function>} gameScriptClass - class needs by object3D
    */
-  run(gameScriptClass = {}) {
+  start(gameScriptClass = {}) {
     // buffer
     let commands = null;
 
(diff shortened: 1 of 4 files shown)


packages/node/src/Game/Thread.js
```diff
@@ -155,153 +164,135 @@ const Parent = class {
...
-    /** @type {Game.Context} */
-    let gameContext = null;
+  /**
+   * Run child process which can be summarize as so:
+   * Listen {@link EVENT} of the parent and wait the gameobject
+   * When gameobject is received launch a {@link Game.Context} and step it over time
+   *
+   * @param {Object<string,Function>} gameScriptClass - class needs by object3D
+   */
+  run(gameScriptClass = {}) {
```
valentinMachado, line 202: start looks like a better name

mathieuLivebardon (approved):
👶

PR #651, opened 2023-05-10
feat(SideBarWidget) add 3DTiles loading feedbacks

valentinMachado:
Add feebacks of the 3DTiles loading (an ui element and some box3 in the scene)

Depends of this [PR](https://github.com/iTowns/itowns/pull/2096)


packages/browser/src/SideBarWidget/SideBarWidget.js
```diff
@@ -193,15 +194,70 @@ export class SideBarWidget {
...
+
+        c3DTilesLayer.addEventListener(
+          itowns.C3DTILES_LAYER_EVENTS.ON_TILE_CONTENT_LOADED,
+          ({ tileContent }) => {
+            if (currentLoadingBox.has(tileContent.tileId)) {
+              this.frame3DPlanar.scene.remove(
+                currentLoadingBox.get(tileContent.tileId)
+              );
+              currentLoadingBox.delete(tileContent.tileId);
+              this.c3DTilesLoadingDomElement.hidden =
```
valentinMachado, line 254: nothing more for this layer could not be true for all of them

valentinMachado:
in order to test this feature you could have to remove the loading screen

clementcolin:
Indeed, loading screen needs to be removed to visualize this feature.

It seems that the UI element disappears when one tile has been loaded in the scene, should it not wait for all tiles to be loaded before disappearing?   
Also, I could not see the 3D elements in the scene that show a tile loading (from what I understand, it should be a cube that uses the bounding box of the tile being loaded)

Does Itowns wait for a tile to be loaded before requesting another ? If not, should we not display multiple 3D object for each loading tile ?

clementcolin:
Yay, it works fine now :) I do not understand how it worked before without the event.tile_requested though.

clementcolin (approved):

PR #659, opened 2023-05-15
646 clean sidebarwidget

valentinMachado:
API has been reworked by deleting the Component folder in @ud-viz/browser src.
THREEUtil has been quite cleaned by removing the module.exports to use instead export way.
SideBarWidget is now call ShowRoom and has been moved in ./examples/assets/js.
Also remove some css from Frame3DBase.css since the change of css importation order introduced by moving SideBarWidget.ss to show_room.css was conflicting (this is why some examples has now to write .style.position = "relative" and .style.zIndex = 2)


clementcolin (approved):
I think that the API rework part could have been done in another PR if someone wanted to discuss this part specifically.

valentinMachado:
> I think that the API rework part could have been done in another PR if someone wanted to discuss this part specifically.

Actually this is more a rework of the sources structure, since SideBarWidget has been moved, src folder was just containing Component folder this is why this layer (Component folder) has been removed. The API per se has not changed (just SideBarWidget is no longer expose).

mathieuLivebardon (approved):
👌

PR #665, opened 2023-05-16
364 edit config files on back end

valentinMachado:
[json editor bundle](https://github.com/josdejong/jsoneditor) has been integrated in examples in debug mode.
Remove SCRIPT_TAG_RELOAD injection and generalize with the term DEBUG_CONTENT (ie injection of script tag reload + json editor).

To test it run an example then press `e` to display the editor you can then select a config file edit it then click save to upload changes on the backend. To see your changes just reload the page.

EDIT: I am not sure this satisfy the need here since it just save some alt-tab ^^


valentinMachado:
The need here seems more to be a kind of real time editor to configure a widget.

PR #671, opened 2023-05-25
clean(examples) EPSG:3946 is define in config

valentinMachado:
EPSG:6946 transformation is define in config


examples/3DTilesPointCloud.html
```diff
@@ -12,27 +12,24 @@
...
         ])
         .then((configs) => {
-          // http://proj4js.org/
-          // define a projection as a string and reference it that way
-          // the definition of the projection should be in config TODO_ISSUE
           udvizBrowser.proj4.default.defs(
-            'EPSG:3946',
-            '+proj=lcc +lat_1=45.25 +lat_2=46.75' +
-              ' +lat_0=46 +lon_0=3 +x_0=1700000 +y_0=5200000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs'
+            configs['crs'][0].name,
```
clementcolin, line 23: We could have make a loop through all crs config to register them (if they are not known by proj4).
Hovewer, it is still fine like that

clementcolin (approved):

clementcolin (approved):

examples/3DTilesPointCloud.html
valentinMachado: I rather that example are doing strict mininum of code to illustrate what is really needed, however I think that extent config should define name of the crs used I am going to implement that

PR #678, opened 2023-05-31
Update AvatarGameShader example setting start position

wesley-petit:
I set the starting position of the current avatar in AvatarGameShader example. I also formated / clean-up the file with Prettier extension.

### DIFF
--- a/examples/AvatarGameShader.html
+++ b/examples/AvatarGameShader.html
@@ -22,7 +22,7 @@
           './assets/config/layer/3DTiles.json',
           './assets/config/layer/base_maps.json',
           './assets/config/layer/elevation.json',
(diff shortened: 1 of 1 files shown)


valentinMachado:
Hum looks like our formatter are conflicting... Could you try to use [eslint-prettier formatter](https://marketplace.visualstudio.com/items?itemName=rvest.vs-code-prettier-eslint) ? If these diff are removed we should then update the documentation to use this extension.

valentinMachado (approved):

valentinMachado:
> My bad, I reformatted the file with the correct extension. I will also update UD-Viz doc on prettier extension.

No worries, actually the doc was not correct [I just did it](https://github.com/VCityTeam/UD-Viz/pull/681)

PR #684, opened 2023-06-01
clean(@ud-viz/browser) remove jquery

valentinMachado:


clementcolin (approved):

clementcolin:
For my knowledge, why do we need to get rid of jquery ? Is XMLhttprequest better in any way ? Should it be the default package to use to get data from an url ?

valentinMachado:
> For my knowledge, why do we need to get rid of jquery ? Is XMLhttprequest better in any way ? Should it be the default package to use to get data from an url ?

It was useless to import all jquery package in @ud-viz/browser to use one function available in the javascript NDK. itowns provide [Fetcher API](http://www.itowns-project.org/itowns/docs/#api/Provider/Fetcher) since we are importing it, we could say that we should use it. But imo using javascript NDK is fine for now, get rif of jquery was the priority.

PR #692, opened 2023-06-20
fix(AudioController): spatialized sounds decompose feature

mathieuLivebardon:


valentinMachado:
This is required to avoid crash ? Since only the position is needed further there was an optimization with 2 less new operation.

mathieuLivebardon:
> This is required to avoid crash ? Since only the position is needed further there was an optimization with 2 less new operation.

The function decompose needs three arguments. Like it was done in [Util.js](https://github.com/VCityTeam/UD-Viz/blob/master/packages/browser/src/Game/External/ScriptTemplate/Component/Util.js#L13)

valentinMachado (approved):

PR #697, opened 2023-07-13
695 change default maxsubdivisionlevel from 3 to 5

valentinMachado:
Change maxSubdivisionlevel from 3 to 5 in order to have the same default value that in itowns.
Also remove this.size attribute in Frame3D since it can be directly get with the web api


mathieuLivebardon (approved):
👌

PR #698, opened 2023-07-13
upgrade(NativeCommandManager) possibility to customize speed

valentinMachado:
possibility to override computeObjectSpeed(Translate/Rotate) where the first argument is the object to move


mathieuLivebardon (approved):
🐞👷

PR #708, opened 2023-08-24
update(ud-viz) itowns 2.40.0

valentinMachado:


package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "ud-viz-monorepo",
   "version": "3.1.1",
-  "lockfileVersion": 3,
+  "lockfileVersion": 2,
```
LorenzoMarnat, line 4: The lockfileVersion was changed to 3 in release 3.1.1, is it a good idea to change it back to 2 ?

valentinMachado: oh I did not pay attention actually I have deleted the old package-lock with `npm run reset` because just typing `npm i` was not working. I think we should stay in 3, I 'll try to fix it.

LorenzoMarnat: lockfileVersion was set to 3 because we used node 18/npm 9 instead of node 16/npm 8

valentinMachado: done

LorenzoMarnat (approved):

mathieuLivebardon (approved):

PR #716, opened 2023-08-28
feature(examples) add context menu copy coord showroom

valentinMachado:
To test right click on the view of the show room example


valentinMachado:
I don't reproduce this behavior. Right clicing one time open the menu and disable itowns controls then left click or right click close the menu and restore itowns controls (but I don't the camera moving..)

mathieuLivebardon (approved):

PR #730, opened 2023-09-07
fix(ud-viz) socket.io dependencies

valentinMachado:


mathieuLivebardon (commented):
I notice you can apply a npm audit fix --force to avoid dependencies vulnerabilities. I think we can use this PR to fix that.

mathieuLivebardon (approved):

PR #746, opened 2023-10-13
Use 2018 buildings

LorenzoMarnat:
Use 2018 buildings instead of 2015 buildings, some triangles are missing in 2015 3DTiles because they were created with an old version of py3dtilers

2015

image

2018

image

(I've created Lyon1 --> Lyon9 2018, those tilesets are available on [3DTiles liris](https://dataset-dl.liris.cnrs.fr/three-d-tiles-lyon-metropolis/2018/))


mathieuLivebardon (approved):
🦫

PR #776, opened 2023-11-30
feat(game_browser_template) different controller

valentinMachado:
game_avatar example can switch between controller
draganddrop_avatar example implement the new controller

there is two type of controller


mathieuLivebardon (approved):
Singleton ?

PR #777, opened 2023-11-30
feat(point_cloud_visualizer) clipping plane

valentinMachado:
Possibility to make a clipping plane visible and transformable thanks a gizmo 
Possibility to enable/disable the clipping


mathieuLivebardon:
TableMemeGIF

PR #781, opened 2023-12-18
Added an equivalent of UD-Viz template.

EricBoix:
Eventually move this content to the UD-Viz-template repository and merge it (with some updates) with UD-Viz-template's content. The reason for this is that JSDoc does not accept the my-demo directory (but only markdown content). And there is no legit place for my-demo into the UD-Viz sources.


mathieuLivebardon:
**CI for this PR failed**:
UD-Viz's _docs_ folder is used by **JSDOC** to put in generated docs code from the various packages in the _./docs/html_ folder, but also to generate static docs with what's in the _./docs/static_ folder. However, **JSDOC** can't read the contents of the _my-demo_ template as a static file (it needs markdown files). Maybe we can create a package or another folder at the root of the project (something like ./templates) to hold the _my-demo_ code?

EricBoix:
Creating a new package within UD-Viz makes no real sense. I would opt for updating https://github.com/VCityTeam/UD-Viz-Template with this content. The difficulty is that there is probably some UD-Viz-Template content to upgrade to this new
ways of creating your own application. @mathieuLivebardon What should we save within UD-Viz-Template ?

EricBoix:
Content moved to https://github.com/VCityTeam/UD-Viz-Template/pull/52.
Closing.

PR #816, opened 2024-05-13
Layer-choice-event 

mathieuLivebardon:
- Layer Choice class extends EventDispatcher
- When Focus  On 3D Tiles function is called, an Event is dispatch at the end of the camera's travel. It returns data which contains focused targetPosition and focused  layer Object

### DIFF
--- a/packages/utils_browser/src/itownsUtil.js
+++ b/packages/utils_browser/src/itownsUtil.js
@@ -78,7 +78,7 @@ export function focusCameraOn(view, controls, targetPos, options = {}) {
           view.controls.dispose();
           view.controls = undefined;
         }
-        resolve();
+        resolve(targetPos);
       }, timeoutDuration);
     } catch (e) {
       reject(e);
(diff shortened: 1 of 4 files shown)


packages/utils_browser/src/itownsUtil.js
```diff
@@ -87,15 +87,15 @@ export function focusCameraOn(view, controls, targetPos, options = {}) {
...
 
 /**
- *
  * Focus a C3DTiles Layer
  *
  * @param {itowns.PlanarView} itownsView - view
  * @param {itowns.C3DTilesLayer} layer - layer to focus
- * @todo this function is used by widget should be contribue or removed
+ * @returns {Promise | null} Promise of the camera focusing on C3DTiles
+ * @todo this function is used by widget should be contribue or remosved
```
LorenzoMarnat, line 95: typos in `contribue` (contributed) and `remosved` (removed)

LorenzoMarnat (changes requested):

LorenzoMarnat (approved):

PR #819, opened 2024-06-10
Space-time shapes

CorentinGaut:
New method to visualise the evolution of the city. We integrate all the versions simultaneously in the city to allow a better understanding of the evolution. 
We implement our method through 4 timelines shapes:
- Vector
- Helix
- Circle
- Parabola

In addition, we implement two ways to position the versions accross the timeline:
- Sequential
- Chronological


mathieuLivebardon (approved):
WinnerWinnerGIF

PR #838, opened 2024-08-28
fix(guided_tour): allow missing layers

LorenzoMarnat:


mathieuLivebardon (approved):
WhatsHisConditionScottForresterGIF

